PR #1, opened 2017-05-01
Add some more number types, minor changes

johnynek:


src/main/scala/com/monovore/decline/Argument.scala
```diff
@@ -5,7 +5,7 @@ import java.nio.file.{InvalidPathException, Path, Paths}
 
 import cats.data.{Validated, ValidatedNel}
 
-trait Argument[A] {
+trait Argument[A] { self =>
```
johnynek: we could seal Argument and give a constructor function to create it. I kind of like that so it discourages people from doing OO designs and putting these things in funny places.

Also, you can add some combinator methods on Argument, which I did, then deleted.

bkirwi: I had `map` on here briefly, IIRC, but ended up dropping it... didn't save much code, and I ended up forgetting to update the `defaultMetavar`.

On the fence about sealing -- I kind of like the fact that folks can create an `Argument[KafkaTopic]` or whatever and have things work out. What kind of misuse are you picturing here?

src/main/scala/com/monovore/decline/Argument.scala
```diff
@@ -21,23 +21,20 @@ object Argument {
...
-      catch { case nfe: NumberFormatException => Validated.invalidNel(s"Invalid integer: $string") }
-
-    override def defaultMetavar: String = "integer"
-  }
+  implicit val readInt: Argument[Int] = readNum("integer")(_.toInt)
+  implicit val readLong: Argument[Long] = readNum("integer")(_.toLong)
+  implicit val readBigInt: Argument[BigInt] = readNum("integer") { bi => BigInt(new java.math.BigInteger(bi)) }
+  implicit val readFloat: Argument[Float] = readNum("floating-point")(_.toFloat)
+  implicit val readDouble: Argument[Double] = readNum("floating-point")(_.toDouble)
+  implicit val readBigDecimal: Argument[BigDecimal] = readNum("floating-point")(BigDecimal(_))
```
bkirwi: `floating-point` feels a bit off for a `BigDecimal`... maybe `s/floating-point/decimal/` for the above three?

src/main/scala/com/monovore/decline/Parser.scala
```diff
@@ -112,7 +112,7 @@ private[decline] object Parser {
     }
   }
 
-  trait Accumulator[+A] {
+  sealed trait Accumulator[+A] {
```
bkirwi: :+1:

src/test/scala/com/monovore/decline/ParseSpec.scala
```diff
@@ -32,6 +32,38 @@ class ParseSpec extends WordSpec with Matchers {
       result should equal("man")
     }
 
+    "work as expected with numeric options" in {
```
bkirwi: :bowing_man:

src/main/scala/com/monovore/decline/Argument.scala
bkirwi: Ah, right, there are floating-points that are not decimals.

bkirwi:
Did the `s/floating-point/decimal/` for `BigDecimal` only, after merging. Thanks for the cleanup &c!

PR #9, opened 2017-07-13
ScalaJS support

bkirwi:
A couple things:
- Do the build changes needed to support scalajs and still build the microsite / etc.
- Split out `Argument[Path]` to a platform-specific file for the JVM.
- Add some code to grab the standard `process.argv` info in `node` and use those when available. (Recent scalajs recognizes normal-looking main methods, but it doesn't seem to thread the arguments through.)


13h3r:
I will take a look this week

core/shared/src/main/scala/com/monovore/decline/CommandApp.scala
```diff
@@ -22,7 +22,7 @@ class CommandApp(command: Command[Unit]) {
   }
 
   def main(args: Array[String]): Unit =
-    command.parse(args) match {
+    command.parse(PlatformApp.ambientArgs getOrElse args) match {
```
johnynek: I'm confused about this one. This is just a method, and it requires `Array[String]` to call the function. It seems weird to me to discard what was passed after being called on a certain platform (I guess JS will have ambientArgs but not scala-jvm or scala-native).

Seems to me this should be higher up. So the caller of this method should call it with ambientArgs.

bkirwi: Hands a little tied here, unfortunately. :disappointed:

This is a `main` method, which is required to take `Array[String]` as a param. ScalaJS learned to recognize regular main methods in the last release, but empirically won't actually pass any arguments to them: it will always pass the empty array. That would make this `CommandApp` sort of useless -- since the whole point is to generate a main that parses args for you -- so we have to go rummaging around in the process state for the args that were actually passed to nodejs.

johnynek:
Okay. Bummer.

core/shared/src/main/scala/com/monovore/decline/CommandApp.scala
13h3r: I agree with @johnynek. During my experience I found that `CommandApp` is not really useful for me. Usually I end up with something like this:
```
val params: Params[XXX] = {
  ...
}
def run(xxx: XXX) = ???
new CommandApp(
      "xxx",
      "xxx",
      params.map(run)
    ).main(Array.empty)
```

core/shared/src/main/scala/com/monovore/decline/Argument.scala
```diff
@@ -10,7 +10,7 @@ trait Argument[A] { self =>
   def read(string: String): ValidatedNel[String, A]
 }
 
-object Argument {
```
bkirwi: Nice catch; fixed!

core/shared/src/main/scala/com/monovore/decline/CommandApp.scala
bkirwi: Huh, interesting. The [intended use](http://ben.kirw.in/decline/usage.html#defining-an-application) of `CommandApp` is something like:

```scala
// in MyApp.scala
object MyApp extends CommandApp(
  name = "xxx",
  header = "xxx",
  main = {
    val params: Opts[XXX] = ???
    def run(xxx: XXX): Unit = ???
    params.map(run)
  }
}
```

...which, on the latest ScalaJS, would generate a script such that you could:

```bash
nodejs js/target/scala-2.11/decline-test-fastopt.js --my-param
```

The generated main method isn't intended to be called by user code... it's there so that the JVM / ScalaJS can use it as the entry point for the application. (I could probably document this better!) Would the above style work for your application, or is there some reason you can't structure your code this way?

PR #10, opened 2017-08-07
cats 1.0.0-MF, scalacheck 1.13.5

tpolecat:
This updates decline to cats 1.0.0-MF, which was trivial … it only required removing the deprecated `CartesianBuilder` syntax. Also updates to ScalaCheck 1.13.5 since 1.13.3 seems to have disappeared.

Many downstream cats libs are publishing milestones depending on 1.0.0-MF so we will all be ready when the final release happens. It would be great to see one from decline!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
 project/*
-target/*
+target/
--- a/README.md
+++ b/README.md
@@ -13,13 +13,13 @@ object HelloWorld extends CommandApp(
   name = "hello-world",
   header = "Says hello!",
   main = {
-    val userOpt = 
+    val userOpt =
       Opts.option[String]("target", help = "Person to greet.")
         .withDefault("world")
-    
+
     val quietOpt = Opts.flag("quiet", help = "Whether to be quiet.").orFalse
 
-    (userOpt |@| quietOpt).map { (user, quiet) => 
+    (userOpt, quietOpt).mapN { (user, quiet) => 
 
       if (quiet) println("...")
       else println(s"Hello $user!")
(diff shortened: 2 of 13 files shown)


bkirwi:
This is great! Thanks in particular for all those applicative-syntax conversions... I wasn't looking forward to hauling those out, but `.mapN` makes it fairly nice.

I'm _slightly_ disinclined to publish a release for M1: arg-parsing libraries tend to be a 'leaf' dependency, and my Maven Central release process is still... uncomfortably manual. Will definitely do it if you're blocked on it, though.

Unrelatedly: cool to see `tuco` picked up a `decline` dep recently! Any feedback?

tpolecat:
It works great! No complaints at all. Oscar recommended it to me.

So yeah we're trying to switch to cats at work and will need decline, but I can build a one-off locally if you're still working on your publishing setup. I see you're using sbt-release … you might also look at sbt-sonatype, which makes it every so slightly less bloody miserable by automating Sonatype's execrable clicky thing. [atto](https://github.com/tpolecat/atto) is a pretty simple build (heh-heh) if you'd like to see an example. (Full disclosure, I have no idea what I'm doing.)

bkirwi:
Yeah, I should definitely switch to prodding Sonatype directly. I'll aim to get a milestone out in the next couple days... will note here when that's up.

tpolecat:
👍

bkirwi:
Aight, `0.4.0-M1` is up on bintray -- I think the easiest resolver for it should be `resolvers += Resolver.jcenterRepo`. (I'll try and get it over to central also ~soon, but not tonight.)

tpolecat:
Brilliant, thanks!

tpolecat:
Works!

PR #28, opened 2018-05-04
Add Metavar[A] type class to provide default metavar names.

non:
This change removes the "defaultMetvar" field from Argument[A], and
adds a new Metavar[A] type class to provide the same functionality.
The API impact of this is minimal: there is an extra type class
instance required when constructing opts, but a low-priority default
Metavar implementation ensures that this is unlikely to be a huge
burden.

The current Metavar implementations were created to match the existing
behavior (e.g. using "integer" for all integer types). I added a bunch
of extra implementations that users might be interested in, but we can
remove those if it seems too noisy.

### DIFF
--- a/core/shared/src/main/scala/com/monovore/decline/Argument.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/Argument.scala
@@ -5,7 +5,6 @@ import java.net.{URI, URISyntaxException}
 import cats.data.{Validated, ValidatedNel}
 
 trait Argument[A] { self =>
-  def defaultMetavar: String
   def read(string: String): ValidatedNel[String, A]
 }
 
@@ -16,16 +15,12 @@ object Argument extends PlatformArguments {
   implicit val readString: Argument[String] = new Argument[String] {
 
     override def read(string: String): ValidatedNel[String, String] = Validated.valid(string)
-
-    override def defaultMetavar: String = "string"
   }
 
   private def readNum[A](typeName: String)(parse: String => A): Argument[A] = new Argument[A] {
     override def read(string: String): ValidatedNel[String, A] =
       try Validated.valid(parse(string))
       catch { case nfe: NumberFormatException => Validated.invalidNel(s"Invalid $typeName: $string") }
-
-    override def defaultMetavar: String = typeName
   }
 
   implicit val readInt: Argument[Int] = readNum("integer")(_.toInt)
@@ -40,7 +35,5 @@ object Argument extends PlatformArguments {
     override def read(string: String): ValidatedNel[String, URI] =
       try { Validated.valid(new URI(string)) }
       catch { case use: URISyntaxException => Validated.invalidNel(s"Invalid URI: $string (${ use.getReason })") }
-
-    override def defaultMetavar: String = "uri"
   }
 }
--- a/core/shared/src/main/scala/com/monovore/decline/Argument.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/Argument.scala
@@ -1,8 +1,7 @@
 package com.monovore.decline
 
 import cats.data.{Validated, ValidatedNel}
-import java.net.{URI, URISyntaxException, URL, MalformedURLException}
-import java.nio.file.{FileSystems, InvalidPathException, Path}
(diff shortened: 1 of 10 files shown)


core/shared/src/main/scala/com/monovore/decline/Argument.scala
```diff
@@ -1,46 +1,88 @@
...
-      catch { case use: URISyntaxException => Validated.invalidNel(s"Invalid URI: $string (${ use.getReason })") }
+  implicit val readUUID: Argument[UUID] =
+    Argument.instance { s =>
+      try { Validated.valid(UUID.fromString(s)) }
+      catch { case _: IllegalArgumentException => Validated.invalidNel(s"Invalid UUID: $s") }
+    }
 
-    override def defaultMetavar: String = "uri"
-  }
+  implicit val readNioPath: Argument[Path] =
```
bkirwi: This will break our friends in scalajs... stuff like this needs to live in the `PlatformArguments` companion.

core/shared/src/main/scala/com/monovore/decline/Argument.scala
```diff
@@ -1,46 +1,88 @@
...
+    instance { s =>
+      try Validated.valid(parse(s))
+      catch { case nfe: NumberFormatException => Validated.invalidNel(s"Invalid $typeName: $s") }
+    }
 
-  private def readNum[A](typeName: String)(parse: String => A): Argument[A] = new Argument[A] {
-    override def read(string: String): ValidatedNel[String, A] =
-      try Validated.valid(parse(string))
-      catch { case nfe: NumberFormatException => Validated.invalidNel(s"Invalid $typeName: $string") }
+  implicit val readBoolean: Argument[Boolean] =
```
bkirwi: I'd originally left this out since it leads to pretty UNIX-unidiomatic CLIs... instead of `cmd --foo true` or `cmd --foo false`, people usually do `cmd --foo` and `cmd --no-foo` or similar.

core/shared/src/main/scala/com/monovore/decline/Argument.scala
```diff
@@ -1,46 +1,88 @@
...
+    }
+
+  implicit val readURI: Argument[URI] =
+    Argument.instance { s =>
+      try { Validated.valid(new URI(s)) }
+      catch { case use: URISyntaxException => Validated.invalidNel(s"Invalid URI: $s (${ use.getReason })") }
+    }
 
-  implicit val readURI: Argument[URI] = new Argument[URI] {
+  implicit val readURL: Argument[URL] =
```
bkirwi: IIRC I left this one out as a lowkey way to push people to use URI instead.

non:
I'm going to wait to get @bkirwi's feedback on this. If he's happy with the changes I'll add tests for all the new arguments and clean things up a bit more.

bkirwi:
> By decoupling this it meant we could potentially derive arguments [...]

This is an interesting idea -- would that be deriving arguments for case classes or similar?

> [...] (or at least, not have to worry about choosing the "wrong" metavar when creating them).

This might be going over my head, since `Argument` and `Metavar` are 1:1 in this PR. Do you have an  example of a situation where we'd want to provide one and not the other, and how that makes things better for the user? (Especially given the default `Metavar` impl, it's not clear why a `Argument` author would be unable to provide an instance... and I'm hesitant to expand the number of concepts in the API.)

johnynek:
I think the challenge is that I thought the intent of metaVar was describe the parameter, not the type (such is he problem with lawless typeclasses ;)

If defaultMetaVar should be the type name, that makes more sense that it can be on argument. Although it does call into question what happens when you mapValidated and similar transforms.

bkirwi:
Yeah, that makes sense. Seems like there are a few things worth doing:
- Add (most of) the new `Argument` instances implemented here.
- Add some documentation for the `Argument` typeclass, including the method semantics and some notes on which `Argument` instances are "intentionally missing" and why and what to do instead.
- Remove the default metavar for `Opts.argument` calls, since the typename doesn't usually provide enough info in that case.

Does that all sound right?

bkirwi:
I think this is closed out with the doc clarifications in #29 and some other followup, but please reopen if we want to do more here!

PR #30, opened 2018-06-18
Misc. updates including refined

travisbrown:
Bumping refined from 0.8 to 0.9 is the only important thing here, but I thought I might as well do the rest while I was here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: scala
 scala:
-  - 2.11.11
-  - 2.12.4
+  - 2.11.12
+  - 2.12.6
 jdk:
   - oraclejdk8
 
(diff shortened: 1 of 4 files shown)


johnynek:
Cool. I need to try refined!

bkirwi:
This is great! Thanks for keeping us up.

bkirwi:
Hmm... it looks like one of these tweaks might have broken sonatype publishing? (Or perhaps that broke on its own in the last few weeks, though we haven't made any other build changes.)

Might be a few days before I have a chance to mess with this again. Are you blocked waiting on the new refined version?

bkirwi:
Never mind, the problem was me! (Had to set up 1.1 for sonatype publishing.) This is now being released as `0.4.2`.

travisbrown:
@bkirwi Sorry, should have flagged that. Thanks for publishing!

PR #36, opened 2018-08-05
Environment variables

### LINKED ISSUE #32: Arguments from environment variables
It would be nice to be able to add the option to parse arguments from an environment variable. Is that something this library would be interested in doing or should it be provided by some other library?

alexjg:
First stab at implementing environment variable parsing. Fixes #32


core/shared/src/main/scala/com/monovore/decline/Help.scala
```diff
@@ -49,11 +49,17 @@ object Help {
       else ("Options and flags:" :: optionsDetail).mkString("\n") :: Nil
     }
 
+    var envVarHelp = {
```
johnynek: Why is this a var?

core/shared/src/main/scala/com/monovore/decline/Help.scala
```diff
@@ -49,11 +49,17 @@ object Help {
...
+      if (envVarNames.isEmpty) Nil
+      else ("Environment Variables:" :: envVarNames.map("    " ++ _)).mkString("\n") :: Nil
+    }
+
     Help(
       errors = Nil,
       prefix = NonEmptyList(parser.name, Nil),
       usage = Usage.fromOpts(parser.options).flatMap { _.show },
-      body = (parser.header :: optionsHelp) ::: commandHelp
+      body = ((parser.header :: optionsHelp) ::: envVarHelp) ::: commandHelp
```
johnynek: I don’t think you need the extra parens here do you? 

Also just one set:
`parser.header :: (optionsHelp ::: envVarHelp ::: commandHelp)` is the most efficient since lists are build from the right to the left.

core/shared/src/main/scala/com/monovore/decline/Help.scala
alexjg: Ha, too much programming in Go.

core/shared/src/test/scala/com/monovore/decline/ParseSpec.scala
```diff
@@ -329,5 +329,25 @@ class ParseSpec extends WordSpec with Matchers with Checkers {
...
+
+      "the variable is not valid" should {
+        "display a suitable error" in {
+          val opts = Opts.envVar[Int]("WHATEVER")
+          val env = Map("WHATEVER" -> "someint")
+          val Invalid(errs) = opts.parse(List(), Map("WHATEVER" -> "invalidint"))
+          errs should equal(List("Error reading WHATEVER from environment: Invalid integer: invalidint"))
+        }
+      }
+    }
```
bkirwi: We should have a test for when the variable's missing entirely, too.

core/shared/src/test/scala/com/monovore/decline/HelpSpec.scala
```diff
@@ -16,16 +17,18 @@ class HelpSpec extends WordSpec with Matchers {
...
             Opts.argument[String]("task")
           }
 
-        (first, second, subcommands).tupled
+        (first, second, third, subcommands).tupled
       }
 
       Help.fromCommand(parser).toString should equal(
-        """Usage: program [--first] [--second <integer>] run
+        """Usage: program [--first] [--second <integer>] --third <integer> run
```
bkirwi: I think this is not the rendering we want: this usage statement makes it looks like `--third` is required, but it's legal to call this program without passing `--third`.

I think changing the added line in `Usage` to match the `Pure` case instead of the `Missing` case should do the right thing, but I haven't tested!

core/shared/src/main/scala/com/monovore/decline/opts.scala
```diff
@@ -159,6 +160,10 @@ object Opts {
   def subcommand[A](name: String, help: String, helpFlag: Boolean = true)(opts: Opts[A]): Opts[A] = {
     Subcommand(Command(name, help, helpFlag)(opts))
   }
+
+  def envVar[A: Argument](varName: String): Opts[A] = EnvVar(varName).mapValidated(raw => {
```
bkirwi: Some thoughts on the interface:
- Slight preference for just `env` over `envVar` (consistent with the other names here, which are single words; still reads fine. Thoughts? Likewise, `varName` should just be `name` for consistency.
- I think it'd be worthwhile capturing a metavar here, since `ENV_VAR=<path>` or whatever seems like useful documentation.
- We might as well capture a `help` string here too. (The only reason `argument` doesn't have one is because it doesn't have a section in the usage output to put it in... but we don't have that problem here!)

core/shared/src/main/scala/com/monovore/decline/Parser.scala
```diff
@@ -319,12 +319,14 @@ private[decline] object Parser {
           .getOrElse(Result.missingArgument)
     }
 
-    case class Subcommand[A](name: String, action: Parser[A]) extends Accumulator[A] {
+    case class Subcommand[A](name: String, action: Parser[A], env: Map[String, String]) extends Accumulator[A] {
```
bkirwi: Ah, hadn't realized we'd need to deal with threading this through -- this looks good!

core/shared/src/main/scala/com/monovore/decline/opts.scala
```diff
@@ -14,7 +14,7 @@ class Command[+A] private[decline](
 
   def showHelp: String = Help.fromCommand(this).toString
 
-  def parse(args: Seq[String]): Either[Help, A] = Parser(this)(args.toList)
+  def parse(args: Seq[String], env: Map[String, String]): Either[Help, A] = Parser(this)(args.toList, env)
```
bkirwi: I'd be tempted to add a default value of `Map.empty` for this, to avoid breaking existing usages.

alexjg:
Everything addressed I think :).

bkirwi:
Huh, looks like a flake in the ScalaJS plugin... haven't seen that before. :disappointed: 

Changes look great... thanks for the rapid followups!

bkirwi:
I'm on vacation for a little bit, so it may be a week or two before I get this pushed to maven central. (Probably as `5.0`!) Will comment here again when that's done!

Issue #35, opened 2018-07-02
Support for effects in CommandApp


kubukoz:
Hi! I just started using decline and it's looking pretty cool.

Did you consider supporting cats-effect style effects? So that there'd be a variant of `CommandApp` like `CommandApp[Task]`, whose `main` would have to return a `Task[Unit]`.

Additionally, if `Opts` had a `Traverse` instance (didn't think about if it's possible yet, but it should be), we could validate them with effects - let's say I have an `Opts[Path]` and want to make sure that the file exists - which is an IO operation, and I'd rather do it in an effect.

Of course I'm willing to help :)

What do you think?

Thanks!

bkirwi:
> Did you consider supporting cats-effect style effects? So that there'd be a variant of CommandApp like CommandApp[Task], whose main would have to return a Task[Unit].

I have! I think it'd make sense to try and integrate with the new [`IOApp`](https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/IOApp.scala) in `cats-effect`, but I haven't looked into it in detail yet.

> Additionally, if Opts had a Traverse instance (didn't think about if it's possible yet, but it should be), we could validate them with effects - let's say I have an Opts[Path] and want to make sure that the file exists - which is an IO operation, and I'd rather do it in an effect.

I'm not totally sure this works... this would lead to code like so:

```scala
val options: Opts[Path] = ???
def validate(path: Path): IO[Path] = ???
val validated: IO[Opts[Path]] = options.traverse(validate)
```

...but it doesn't make much sense to have `IO` on the _outside_; you'd need to run the validation before you even got your hands on the parser!

OTOH, I think there are other ways to get similarly-nice behaviour. I've noticed that `Nested[Opts, IO, ?]` has a pretty nice alternative instance... you get effectively the same behaviour as for regular `Opts`, but with effects that also compose in the obvious way. I haven't had a chance to try this out in real life, though.

bkirwi:
Two points I forgot:
- Really glad you like the libray, and thanks for the suggestion!
- I had figured I'd get around to this problem when I had a couple largish, IO-driven application to play around with: build up a set of utils alongside the code, and then port them here. I haven't got around to this yet... but I'd be delighted to have contributions of that shape!

kubukoz:
[...]

PR #43, opened 2018-10-12
Better doc for parsing in main method

bkirwi:
Added an example of a main method to the doc -- `CommandApp` isn't for everyone, and this should be a good starting point for other uses.

cc @DougC - this came up in #35.


johnynek:
Would be nice to have an IO example in there.

I agree that currently you don’t get much win by having integration but a copy paste example with IOApp would be nice.

PR #45, opened 2018-11-08
Add support for Java 8 time data types

alonsodomin:
An additional module providing with some default implicits for parsing arguments using Java 8 time API. The module is usable in ScalaJS thanks the `scala-java-time` library.

I also took the liberty to rewrite the `ArgumentSpec` and base it in Discipline + ScalaCheck to remove code duplication.


bkirwi:
hey, thanks for this! overall seems solid and thoroughly tested.

will have more time to look at this later, but some quick thoughts:
- i'd prefer to just have these implicits in the main project instead of
new "time" project... a better experience for jvm users, which is the
common case.
- the metavars should be lowercase-hyphenated for consistency.
- the metavar names should be chosen to be as suggestive for the cli user
as possible / usable even if you're unfamiliar with java.time. i'd consider
<iso-date-time> over <instant>, for example.

thoughts?

alonsodomin:
All very good comments @bkirwi, should admit that initially I thought about putting all this in the core module there may be a few reasons why it could belong in its own separate module:

 * Support for Scala 2.11 and Java 7. Since the time library has been added in Java 8, it's is not available in Java 7 JVMs. It could still be usable in Scala 2.11 if its run in a modern JVM but not in the latter... Not sure if you want to support that combination though...
 * Ability to set date-time formats other than the defaults. Meaning that users that are happy using the standard ISO formats in the command line would only need to do a single `com.monovore.decline.time._` whilst the ones that prefer to use other formats could just declare their own implicits (i.e.: `implicit val myDateArg: Argument[LocalDate] = time.localDateWithFormatter(DateTimeFormatter.ofPattern("dd/MM/yy"))`)

Event myself is not too convinced about that second point but well... What do you think about that? Some overthinking in my side maybe?

I'll address all the other points ASAP

bkirwi:
Been thinking about this a little bit!

I'm nervous about the JVM-7-support argument, since 7's so long past end of life, and as you say 2.11 works fine on 8 Java 8 JVMs. Really hard to quantify, though...

Supporting non-default arguments is an interesting argument. So far we've taken the approach of being opinionated and trying to have one impl for all argument types -- though I understand that time formats are a bit of a harder case than eg. URIs. (I think we can still provide defaults in the companion, though -- IIUC explicit imports will still override implicits in the companion, so folks can get that behaviour if they want it.)

I think my preference is still for merging it in to the main artifact. We can leave it in a separate package for now... that should avoid problems for Java 7 laggards, and keep some other options open. We can always haul the implicits to the main companion object in a followup PR. WDYT?

alonsodomin:
sorry for the delay coming back to this. I merged the `time` module with core and I also tested it with a Spark application running in a Java 7 cluster. All seems to work well as long as the code is no near the `time` package, which is great.

bkirwi:
Great! Really appreciate the followup on an already very thoughtful PR.

Merging now; I'll aim to get a new release up on Maven Central within the next day or so.

bkirwi:
Released as `0.6.0`!

Next up is to update the docs: recommend 0.6.0 and add something about `time` to `arguments.md`. I'm happy to take, unless you'd rather grab it!

alonsodomin:
Great, yeah, there should be some hint at the new java time support in the docs, that should be easy and quick.

PR #46, opened 2018-12-03
Update docs

alonsodomin:
So this is my attempt at updating the docs.

I added two new sections, one in **Usage** for showing the most basic usage relying in the ISO 8601 format. The other section is in `Arguments` for showing how to customize a date format.

Let me know what you think.

### DIFF
--- a/doc/src/main/tut/arguments.md
+++ b/doc/src/main/tut/arguments.md
@@ -78,6 +78,34 @@ implicit val configArgument: Argument[Config] = new Argument[Config] {
 val config = Opts.option[Config]("config", "Specify an additional config.")
 ```
 
+## Custom time-based arguments
+
+As mentioned in the Usage page, `decline` has out-of-the-box support for Java 8 time library by importing the contents
+of the `com.monovore.decline.time` package. But this usage is limited to the standard ISO 8601 formats and some people
+would like to have a more human-friendly format. Fortunately we provide with utility functions that you can use to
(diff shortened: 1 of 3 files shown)


doc/src/main/tut/arguments.md
```diff
@@ -78,6 +78,34 @@ implicit val configArgument: Argument[Config] = new Argument[Config] {
...
+```tut:silent
+import java.time._
+import java.time.format.DateTimeFormatter
+import com.monovore.decline.{time => timeargs}
+```
+
+With the previous we can now move on to define the argument parsers:
+
+_**NOTE:** The renaming of the `time` package to `timeargs` is not really necessary but it is recommended just to remove
+ambiguities and potential name collisions._
```
bkirwi: Hmm... if we're going to always suggest users rename the package, that suggests to me that we should think about expose things differently. (Maybe the builders belong on another object?)

bkirwi:
Thanks for this! This seems very thorough.

I've got a little reworking of the docs to do, so I may end up shuffling these around a bit. (Currently the usage guide doesn't talk about supported argument types at all!) Will cc you on the followup.

doc/src/main/tut/arguments.md
alonsodomin: Yeah, I do agree... I came up with that suggestion when playing a little bit with the library after the fact... sorry for not pointing that out while the merge of that feature was still open...

PR #48, opened 2018-12-05
Fix a flaky test!

bkirwi:
It seems like we can generate timezones that we can't parse! Did an exhaustive test... this seems to be the only one that causes trouble.


bkirwi:
cc @alonsodomin

PR #49, opened 2018-12-08
added a `Monoid` instance for `Opts[A]`

kailuowang:
I can't think of any reason not to provide this instance implicitly, but I might miss something.

### DIFF
--- a/core/shared/src/main/scala/com/monovore/decline/opts.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/opts.scala
@@ -1,6 +1,6 @@
 package com.monovore.decline
 
-import cats.Alternative
+import cats.{Alternative, Monoid}
 import cats.data.{NonEmptyList, Validated, ValidatedNel}
 
 /** A top-level argument parser, with all the info necessary to parse a full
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/com/monovore/decline/opts.scala
```diff
@@ -99,6 +99,8 @@ object Opts {
       override def combineK[A](x: Opts[A], y: Opts[A]): Opts[A] = Opts.OrElse(x, y)
     }
 
+  implicit def monoid[A]: Monoid[Opts[A]] = alternative.algebra[A]
```
johnynek: The shame of it is that this isn't a low priority default if `Alternative[F]` exists...

but anyway, this looks good to me.

bkirwi:
For `Alternative`s, I'm never sure whether to provide this instance or the one that goes via `Applicative`...

In this case, though, this seems both slightly more useful (eg. combining a list of commands) and a bit more direct, so I'm +1 if y'all are into it.

kailuowang:
It seems there is no objections. shall we merge this?

bkirwi:
Merged! I'll get a point release out in a bit.

kailuowang:
Thanks!  @bkirwi

bkirwi:
Thanks for the PR!

`0.6.2` is released with this patch, and should be available from central within a couple hours.

PR #51, opened 2019-01-14
Add UUID support in Argument

BardurArantsson:


bkirwi:
Hey, thanks for this! Great addition... I'm surprised I haven't needed this myself yet.

I'm not religious about coverage -- but adding the basic test we use for arguments (see `ArgumentSuite` or `JavaTimeSuite`) is pretty quick and often finds bugs, so it seems worthwhile here. Let me know if you run into trouble with it.

bkirwi:
Cool, looks great... thanks for the follow-up! I'll get that merged soon.

PR #54, opened 2019-04-16
pass default values to `withDefault` by name

mtomko:
Until parsing happens within the context of an effect, it would be nice if my default values can be evaluated lazily.

### DIFF
--- a/core/shared/src/main/scala/com/monovore/decline/opts.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/opts.scala
@@ -51,7 +51,7 @@ sealed trait Opts[+A] {
 
   def orElse[A0 >: A](other: Opts[A0]): Opts[A0] = Opts.OrElse(this, other)
 
-  def withDefault[A0 >: A](default: A0): Opts[A0] =
+  def withDefault[A0 >: A](default: => A0): Opts[A0] =
     this orElse Opts.apply(default)
 
   def orNone: Opts[Option[A]] =
--- a/core/shared/src/test/scala/com/monovore/decline/ParseSpec.scala
+++ b/core/shared/src/test/scala/com/monovore/decline/ParseSpec.scala
@@ -330,6 +330,16 @@ class ParseSpec extends WordSpec with Matchers with Checkers {
       }
     }
 
+    "lazily evaluate defaults" in {
+      var count = 0
+      val a = Opts.option[Int]("num", "a number").withDefault {
+        count += 1
+        count
+      }
+
+      count should be (0)
+    }
+
     "read from the environment" when {
       "the variable is present" should {
         "read the variable" in {


mtomko:
I guess you're right that it probably would break binary compatibility, which probably means this is a non-starter. 

I'll play around with my use case making it an `Opts[IO[Int]]`, but decline doesn't provide an implicit `Argument[IO]` (nor for `Argument[cats.Eval]`). Having `IO` there just to handle the case when a value isn't provided feels a bit heavy-handed, but I guess it's maybe more honest overall.

Regarding effects in general, I wasn't arguing that it should, only noting that an existing issue (https://github.com/bkirwi/decline/issues/35) raised the question; I hadn't fully digested the conversation yet when I mentioned it but I see now that you and others expressed some skepticism of that idea in the aforementioned issue. I guess the advantage to considering effects that is that things like gathering defaults or other options from the environment is impure; if you end up with lots of `Opts[IO[_]]` you'll have extra work to map them together at the end. 

The other possibility would be to wrap the whole command construction and parsing in, eg, an IO:

```scala
private[this] def unsafeCommand: Command[Config] = // constructing this is potentially impure

def parse(args: List[String]): IO[Either[Help, Config]] = IO(unsafeCommand.parse(args).flatMap {
  case Left(help) => // handle help
  case Right(config) => // run with config
}
```
 
This feels weird to me because `IO` already has its own error handling capability, so the nested `IO` and `Either` feels a touch redundant, unless you treat the `IO` as catching any other unexpected exception.

mtomko:
It's late and I might be missing something obvious, but it seems like defining an `Argument[IO[A]]` will be tricky because the `read` method will be hard to implement:
```scala
trait Argument[A] { self =>

  /**
   * Attempt to parse a single command-line argument: given an argument, this returns either
   * the parsed value or a message explaining the failure.
   */
  def read(string: String): ValidatedNel[String, A]
```
we can't get the `ValidatedNel` without evaluating the `IO`, unless I'm missing something obvious; or would we just assume that it's valid and encapsulate the error in the `IO`?

johnynek:
I guess it depends on how you want to validate. If you need some effect to validate you are right, you are out of luck. I haven’t had a need for that case.

bkirwi:
Yeah, I agree that `Argument[IO[_]]` is probably a dead end, for exactly the reasons you mention. `Opts[IO[_]]` is a fine pattern, though... you can always

```scala
Opts.argument[Path]("path").map(IOActions.checkExists)
```

or something like that. `Nested[Opts, IO]` has a nice `Alternative` instance if you're going down that road... behaves exactly as you'd want and saves some boilerplate.

A big hesitation with laziness here is that it's against the usual Cats style... where things tend to be either strict or uses laziness explicitly with `Eval`. I'd be more enthused about an explicit `withDefaultEval` or something I think. (@johnynek as Cats expert... does that sound right?) I guess I'm a litttle surprised the need arises, though... running into default values that are expensive to compute?

(The one place I _have_ put laziness is in `Opts.apply`, of course, as you're taking advantage of there! Did this so folks who aren't using an effect system could still have zero-argument (sub)commands. Does mean you have the escape hatch of doing `arg orElse Opts { myExpensiveDefault }` too at least...)

Also: thanks for the contribution! Hope you're enjoying the library overall...

mtomko:
I see, I can get the same behavior by doing:

```scala
Opts.argument[Path]("path").map(Eval.later).withDefault(...)
```

That's good enough for me. I'm not well versed in `Alternative` but I'll look up how that works. I'm going to close this PR. Thanks for your input everyone, and thanks for a really nice library!

PR #57, opened 2019-05-04
fix travis build error. use openjdk8 instead of oraclejdk8

xuwei-k:


bkirwi:
Thanks for this! And sorry for the slow turnaround; on holiday.

PR #58, opened 2019-05-04
bump versions

SethTisue:
it would be helpful to the Scala community build to have this
merged (because of the kind-projector org change)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -66,7 +66,7 @@ lazy val root =
 lazy val decline =
   crossProject(JSPlatform, JVMPlatform).in(file("core"))
     .settings(defaultSettings)
-    .settings(addCompilerPlugin("org.spire-math" %% "kind-projector" % "0.9.8" cross CrossVersion.binary))
+    .settings(addCompilerPlugin("org.typelevel" %% "kind-projector" % "0.10.0" cross CrossVersion.binary))
     .settings(
       name := "decline",
       description := "Composable command-line parsing for Scala",
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.2.7
+sbt.version=1.2.8


SethTisue:
looks like your `.travis.yml` needs some work, or perhaps Travis-CI is having transient trouble?

SethTisue:
also note that this may not be mergeable unless you also move to Scala 2.13.0-RC1 and cats 2.0.0-M1

bkirwi:
👋thanks!

this config was also touched in the just-merged #56. i do think going to 2.13.0-RC1 makes sense... gotta do it anyways, and the less conditionals in sbt config the better. not clear to me why i'd need to go to cats 2 at the same time... any context? cats is our ~only public dep so i've been a bit more conservative with updates there, though i do expect the upgrade to be mostly painless.

> looks like your .travis.yml needs some work, or perhaps Travis-CI is having transient trouble?

seems that travis was having trouble with the oraclejdk8 builds... rebasing on master ought to fix.

SethTisue:
there was a ripple effect where ScalaCheck only published 1.14, not 1.13, for 2.13.0-RC1. And cats isn't only using ScalaCheck for testing internally, but is also publishing artifacts that have ScalaCheck as a dependency (for other people to use to do _their_ testing with). so cats decided to move to ScalaCheck 1.14 which required they move to 2.0, since 1.13 and 1.14 aren't either binary or source compatible.

The good news is that cats 2 doesn't have a lot of _other_ changes in it, it's not some big radical overhaul.

The bad news is that cats 2 is still in milestones. But my understanding is that it will go into final release not long after Scala 2.13.0 is published. That's my understanding, the cats maintainers would be able to confirm and/or explain further.

SethTisue:
anyway, I think this PR is no longer necessary after @xuwei-k's 0e15fb6f86a11055cfca16a469b8a8c9f73557f0

bkirwi:
ah, that all makes sense... thank you!

PR #60, opened 2019-05-12
Scala 2.13.0

xuwei-k:


Krever:
@bkirwi do you think you could merge and release this?

bkirwi:
Hmm... I've been holding out on a merge because cats, our main dependency,
doesn't have a stable release on 2.13 yet. (And the current milestone seems
likely to break bincompat.) Would you still want a release under those
conditions?

Krever:
This sounds reasonable. I already rebased this PR and published locally and I don't plan to release anytime soon so form me its fine for at least week or two. Do you know when the stable cats is planned?

kailuowang:
> holding out on a merge because cats, our main dependency, doesn't have a stable release on 2.13 yet. (And the current milestone seems likely to break bincompat.)

That's correct. Although, a milestone deline release will be helpful...(at least I can use it. :) )

bkirwi:
Aight, I can milestone if it'll help folks... will comment here when that's done.

bkirwi:
Released as `0.7.0-M0`... holler if you run into any issues.

kailuowang:
Thanks!

PR #61, opened 2019-05-20
Mention `tupled` for combining Opts

BenHutchisonSeek:
When I was learning Decline, this was a key question not covered by the existing docs that I spent a fair bit of time figuring out; want to ease the way for the next person trying out Decline


bkirwi:
Sure, seems good! It's a pain that `cats` doesn't have any docs we can link to here...

PR #66, opened 2019-07-04
Proposal: Support cats-effect IOApp

### LINKED ISSUE #35: Support for effects in CommandApp

alonsodomin:
This is a proposal for supporting `IO`-based applications built with `cats-effect`, main changes are:
* Adds a new `CommandIOApp` class which needs to be extended
* Replaces the `HelpFlag` in the `Opts` applicative by an `Abort` one. This is inspired by the Haskell's `optparse-applicative` and allows for unifying aborting execution due to a `--help` flag or any other user's custom one (i.e. `--version`, as it can be seen in the `CommandIOApp`).

Let me know what your thougts are on this, I assume that we may need to give some more work to this to be mergeable though... (i.e. a test of this...)

Closes #35


bkirwi:
Hey, neat! As mentioned on the issue, this is a pretty appealing feature... and I appreciate the included docs etc.

Two overall, related comments up front:
- Changing the signature of `command.parse` is a breaking and easy-to-miss change for a lot of users. Common for users to do `cmd.parse(args).match { case Left(x) => System.err.println(x.toString); sys.exit(1) }` or similar, which will give the wrong exit code / print stream now. I think the existing `CommandApp` would currently have the wrong behaviour here too!
- Having `class CommandIOApp(command: Command[A])` with a method `def execute(a: A): IO[ExitCode]` is functionally equivalent to defining `class CommandIOApp(command: Command[IO[ExitCode]])` for our purposes. It's not clear we gain anything by the extra API surface.

I'm a little bit loathe to bake in a new effect type to `Opts` directly... and it seems like if we took the `class CommandIOApp(command: Command[IO[ExitCode]])` approach we wouldn't need to, because it would be easy to define a `showVersion: IO[ExitCode]` that printed directly. Would you consider that approach?

alonsodomin:
Yeah, this approach is quite intrussive as it requires the modification of the applicative and others parts that act as the internal machinery. I guess I'm strongly influenced by the usage of `optparse-applicative` in which the common thing is to parse the command line into a data structure and then _interpret_ it later (hence the abstract `execute` method).

I'll give it another go using the `Command[IO[ExitCode]]` approach to see how it looks like.

alonsodomin:
Now I remember why I chose to have a `CommandIOApp` with an abstract `execute` method, the `IOApp` from `cats-effect` provides with some implicits (i.e. `Timer[IO]`, `ContextShift[IO]`) which are required to implement useful concurrent `IO` operations. By extending the `CommandIOApp` those implicits are in scope and providing actions that fulfil al `A => IO[ExitCode]` is trivial for the simple cases (of course, async operations still will require their own instances of those implicits.

Asking for a `Command[IO[ExitCode]]` (or a `Opts[IO[ExitCode]]`) in the class constructor means that the implicits are not in scope and to be able to have one of those, users will need to do something in the likes of the following:

```scala
val myCliOpts: Opts[MyCliOpts] = ???
def myCliInterpreter(opts: MyCliOpts)(implicit timer: Timer[IO]): IO[ExitCode] = ???

myCliOpts.map { opts =>
  val executionContext: ExecutionContext = ???
  implicit val timer = IO.timer(executionContext)
  ... // other implicits
  myCliInterpreter(opts)
}
```

However with the `CommandIOApp`, `myCliInterpreter` is the `execute` method, which doesn't need additional implicit arguments. The implicit arguments will be needed anyway in case the user is interested in define different interpreters in different files/modules. But I find the redefiintion of the defaults in the `map` combinator a bit annoying, since it makes the trivial case no so trivial anymore.

Any idea on how to circumbent this?

johnynek:
My 2 cents are: decline-effect is a nice module to have, but I think it can and should be implemented without any changes to decline core.

bkirwi:
> I guess I'm strongly influenced by the usage of optparse-applicative in which the common thing is to parse the command line into a data structure and then interpret it later (hence the abstract execute method).

That's definitely a common pattern, but working with the effect type directly is also pretty common (especially with subcommands) and I think it's important to support both. (And IME the optparse-applicative style is easier to implement on the "direct" style than the reverse.)

Well spotted on the `ContextShift` / etc. issue... had not occured to me! It does make the class constructor approach less appealing. I guess the alternative is to use the `IOApp` style directly:

```scala
object MyApp extends CommandIOApp {
  def command: Command[IO[ExitCode]] = ???
}
```

Main downside of that pattern is that you can't overload the constructor like `CommandApp` does. We could just bake in the `version` behaviour:

```scala
object MyApp extends CommandIOApp {
  def version = "1.0" 
  def command: Command[IO[ExitCode]] = ???
}
```

Alternately we could just stick a helper on the companion object and tell users to mix it in if they want it:

```scala
object CommandIOApp {
  def printVersion(v: String): Opts[IO[ExitCode]] = ???
}
```

The second one is a bit less magical... but `decline` overall tries pretty hard to implement CLIs that are POSIX / [GNU compliant](https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html) and those guidelines do say every command should have a `version` flag. Maybe the right compromise is to do ~both:
- Have `CommandIOApp` which bakes in the `def version` behaviour.
- Factor out most of the logic to helper methods on `CommandIOApp`, so folks who want to mix and match can do that easily.

Thoughts?

alonsodomin:
What about if the version would be passed as a constructor argument...?

```scala
abstract class CommandIOApp(version: String = "") {
  def command: Command[IO[ExitCode]]
}
```

I'll give a try to some of these ideas over the weekend.

alonsodomin:
Created pull request #67 for doing the same with minimal changes (just the additional `decline-effect` module).

I'll close this one, we can continue the conversation over there.

PR #67, opened 2019-07-06
Support cats-effect IOApp

### LINKED ISSUE #35: Support for effects in CommandApp

alonsodomin:
Another take into supporting effectul applications via `cats-effect`, now without any changes to the `core` module and using an approach as suggested in #66.

In this take, the `CommandIOApp` takes the constructor arguments needed to build a `Command[IO[ExitCode]]` that supports a version flag and asks the user to provide with an implementation of `commandOpts: Opts[IO[ExitCode]]` in which the user can take the advange of the _runtime_ provided by the cats-effect `IOApp` to implement their `IO`s.

Obviously, this is open for suggestions. The fact that the `CommandIOApp` loses the ability of being instantiated by an `Command[IO[ExitCode]]` as the `CommandApp` can do, can be considered a lack of consistency, although it could also be a good trade off...

Let me know what you think.

Closes #35


effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  override final def run(args: List[String]): IO[ExitCode] = {
+    def printHelp(help: Help): IO[ExitCode] =
+      IO(System.err.println(help)).as {
+        if (help.errors.size > 0) ExitCode.Error
+        else ExitCode.Success
+      }
+
+    for {
```
johnynek: When ambient args are set we totally ignore args here? That seems strange to me. Can we comment why?

alonsodomin: Yes, that line will ignore `args` when `ambientArgs` is not a `None`. If I am correct, `ambientArgs` should be a `Some` only when running in a NodeJS process, which made me implement the test only in the JVM because I was not using real command line arguments.

This is just a replication of the same behaviour in `core` for the `CommandApp`. It could also be safe to assume that, since `IOApp` is already cross-platform, we don't need the `ambientArgs` trick anymore.

Shall I get rid of them and rely on `args` only?

bkirwi: Yep, you understand the idea right.

I actually don't know if `IOApp` does the right thing here --worth testing! (I can't see code that would do it, but I might be missing out.) If testing is too painful, we can probably skip -- this code at least shouldn't do any harm.

(If `IOApp` is not doing the right thing, that might be a nice upstream contribution, but we'd want to hang onto it in the meantime.)

doc/src/main/tut/effect.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+Now we need to build our parser composing the individual elements for each of the components. So the `ps` subcommand
+will look like the following:
+
+```tut:book
+val showProcessesOpts: Opts[ShowProcesses] = Opts.subcommand("ps", "Lists docker processes running!") {
+  Opts.flag("all", "Whether to show all running processes.", short = "a")
+    .orNone
+    .map(_.fold(false)(_ => true))
```
bkirwi: `.orFalse` would be more idiomatic here.

effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
```diff
@@ -0,0 +1,42 @@
+package com.monovore.decline.effect
+
+import cats.effect.{IO, IOApp, ExitCode}
+import cats.implicits._
+
+import com.monovore.decline._
+
+abstract class CommandIOApp(name: String, header: String, helpFlag: Boolean = true, version: String = "") extends IOApp {
+
+  def commandOpts: Opts[IO[ExitCode]]
```
bkirwi: I'd be tempted to name this just `opts` or `main`.

effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  private[this] val command: Command[IO[ExitCode]] = {
```
alonsodomin: Now that I look at this with fresh eyes, this `val` is a code smell since it is depending on a `def` and could cause initialization order errors. It should be either a `lazy val` or a `def`, the former giving just the benefit of being synchronized... which I doubt is of any gain in here... any strong thoughts on that?

effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
alonsodomin: According to ScalaJS docs, there should be no need to do any tricks to get the command line arguments when running a ScalaJS process (their example HelloWorld takes a `def main(args: Array[String]): Unit`). But I keener of keeping the same code that is right now in `core` and potentially fix this in a PR that targets both modules.

effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
alonsodomin: `main` it's ok for me

alonsodomin:
The `CommandIOApp.FromCommand` it's not a bad solution, but kinda feel that it's probably better to leave for a later PR... it's quite likely that users may not find a reason to use it...

alonsodomin:
took the liberty of adding `sbt makeMicrosite` to test the documentation during the build and found out that it breaks with Scala 2.13.0-M5 because of a deprecated implicit conversion

effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
bkirwi: Slight preference for `def`.

effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
bkirwi: Yeah, historically they would just call `main` with an empty array. (Sketchy!) Maybe they've fixed it, but I'd want to test to be sure.

+1 to keeping the existing pattern and following up.

alonsodomin:
yeah, this should be ready to merge.

regarding the microsite stuff: apart from removing the fatal warnings during the execution of `tut`; had to split one of the code blocks into two, for some reason the Scala repl for 2.13 didn´t like seeing a comment in there

bkirwi:
Cool!

Going to try and get this out along with the 2.13 milestone.

bkirwi:
(Out under `0.7.0-M0` if you want to try it.)

PR #69, opened 2019-07-11
Fix scalatest deprecation compile error

Taig:
Waiting for CI and will make subsequent changes if necessary


bkirwi:
...that was fast!

PR #70, opened 2019-08-22
Proposal: Enumeratum Support

alonsodomin:
I find quite useful to be able to define enumerated values in my applications and, until Scala 3 arrives, the best support for enumerated values comes via the `enumeratum` library.

This PR adds another module to decline bringing an integration with that library so the cases for a given enumerated type can be used as command line arguments.

PS: Took the liberty to add an instance for `Argument[Short]` since it looked to me that the fact that it was missing may have been overlooked. Still, there is possibility here to add also instances for `Argument[Char]` and `Argument[Byte]` so enums backed by `Char` or `Byte` can also be used.

### DIFF
--- a/enumeratum/src/main/scala/com/monovore/decline/enumeratum/ValueEnumArgument.scala
+++ b/enumeratum/src/main/scala/com/monovore/decline/enumeratum/ValueEnumArgument.scala
@@ -5,15 +5,12 @@ import cats.data.{Validated, ValidatedNel}
 
 import _root_.enumeratum.values._
 
-import scala.reflect.ClassTag
-
 private[enumeratum] final class ValueEnumArgument[A, Entry <: ValueEnumEntry[A]](
     enum: ValueEnum[A, Entry],
-    baseArgument: Argument[A],
-    ct: ClassTag[Entry]
+    baseArgument: Argument[A]
   ) extends Argument[Entry] {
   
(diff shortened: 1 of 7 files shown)


enumeratum/src/main/scala/com/monovore/decline/enumeratum/ValueEnumArgument.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import scala.reflect.ClassTag
+
+private[enumeratum] final class ValueEnumArgument[A, Entry <: ValueEnumEntry[A]](
+    enum: ValueEnum[A, Entry],
+    baseArgument: Argument[A],
+    ct: ClassTag[Entry]
+  ) extends Argument[Entry] {
+  
+  override def defaultMetavar: String = ct.runtimeClass.getSimpleName().toLowerCase()
```
bkirwi: I'd like to avoid a dependency on the runtime class name... could we use the `baseArgument.defaultMetavar` here and maybe just something generic like `value` below?

alonsodomin:
great, glad that you find the contribution interesting, I address the comment regarding the `defaultMetavar` and added `Argument` instances for `Char` and `Byte` to give a broader support to the value enums.

bkirwi:
Changes look good... thanks again for the contribution! I'm going to try and get a release out along with the new cats RC soon as well.

PR #87, opened 2019-09-18
Doc rewrite

bkirwi:
Two things:
- Rewrite and overall condense the docs.
- Use `scalafmt`. (Overall dunno if it's an improvement, but consistency seems worth it.)


bkirwi:
(Implemented the different parts of this in other PRs.)

PR #93, opened 2019-09-23
Parameterize library version in documentation

alonsodomin:
First of all, thanks for the shout out on Twitter, really appreciated it.

I was checking the new docs, which are way more accessible than the previus version, and noticed some inconsistencies with the version numbers on the extension modules. Since you have ported them to Mdoc, we can pass values from the `build.sbt` file into the docs to solve that.

So this is a tiny PR just adding that.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,6 @@
 import ReleaseTransformations._
 import sbtcrossproject.{crossProject, CrossType}
+import microsites._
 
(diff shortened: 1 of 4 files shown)


bkirwi:
Cool, thanks, I'm glad!

And this looks good... nice catch! I think we might want to decouple this from the current version (if we do a milestone release or whatever) but easy enough to cross that bridge when we get there.

Issue #31, opened 2018-06-19
Set up mima


johnynek:
https://github.com/lightbend/migration-manager#usage

would be nice to make sure we don't let compatibility issues creep in.

bkirwi:
:+1: -- this is getting stable enough that we could think about eg. blocking CI on it.

PR #109, opened 2019-11-07
Substitute cats.effect.IO with polymorphic effect type in CommandIOApp

### LINKED ISSUE #101: Support ZIO's effect types and runtime system
As a user, I would expect to be able to use ZIO's runtime system, and use its effect types, like Task.
I think these could be done with two improvements:

- Separating CommandIOApp abstract class definition (with its abstract main method) from the wiring to the cats-effect runtime system (IOApp), and letting the abstract method be polymorphic in the effect type.
- Add two implementations of this abstract class: one which is identical to how it looks now, for the cats-effect runtime system, and one for the ZIO runtime system (maybe in another module, or just depending on ZIO in the effect module?)

With this approach, even Monix could be supported by creating a third implementation based on his own runtime system (TaskApp). Some insights on why each IO monad implementation benefits from his own runtime system can be found [here](https://typelevel.org/cats-effect/datatypes/ioapp.html#why-is-it-specialized-for-io).

Let me know what you think about this and if you are interested in supporting ZIO at all, if yes I could try to sketch a PR.

andreamarcolin:
This PR aims to support using ZIO (and even Monix) with CommandIOApp in a straightforward way.

CommandIOApp now accepts an F[_] type parameter with implicit evidence of Effect[F], which is provided out of the box and automatically imported by cats-effect for IO, while for ZIO and Monix it need to be imported from the respective interop packages.

Its main abstract method is now polymorphic in F and can return any Opts[F[_]].
The behavior of CommandIOApp is maintained by mapping the success case to ExitCode.Success, and handling error (via ApplicativeError) by discarding the error informations and returning ExitCode.Error.

See the updated PureHelloWorld for an example.

Resolves #101

### DIFF
--- a/effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
+++ b/effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
@@ -1,52 +1,70 @@
 package com.monovore.decline.effect
 
-import cats.effect.implicits._
-import cats.effect.{Effect, ExitCode, IO, IOApp}
+import cats.effect.{ExitCode, IO, IOApp, Sync}
 import cats.implicits._
+
 import com.monovore.decline._
 
 import scala.language.higherKinds
 
-abstract class CommandIOApp[F[_]: Effect](
+abstract class CommandIOApp(
     name: String,
     header: String,
     helpFlag: Boolean = true,
     version: String = ""
 ) extends IOApp {
 
-  def main: Opts[F[ExitCode]]
-
-  private[this] def command: Command[IO[ExitCode]] = {
-    val mainCommand: Opts[IO[ExitCode]] =
-      main.map(_.toIO)
-
-    val showVersion: Opts[IO[ExitCode]] = {
-      if (version.isEmpty) Opts.never
-      else {
-        val flag = Opts.flag(
-          long = "version",
-          short = "v",
-          help = "Print the version number and exit.",
-          visibility = Visibility.Partial
-        )
-        flag.as(IO(System.out.println(version)).as(ExitCode.Success))
-      }
-    }
-
-    Command(name, header, helpFlag)(showVersion orElse mainCommand)
-  }
+  def main: Opts[IO[ExitCode]]
 
   override final def run(args: List[String]): IO[ExitCode] = {
-    def printHelp(help: Help): IO[ExitCode] =
-      IO(System.err.println(help)).as {
-        if (help.errors.nonEmpty) ExitCode.Error
-        else ExitCode.Success
-      }
+    import CommandIOApp._
+
+    val opts = if (version.isEmpty) main else addVersionFlag(main)(version)
+    val command = Command(name, header, helpFlag)(opts)
 
     for {
       parseResult <- IO(command.parse(PlatformApp.ambientArgs getOrElse args, sys.env))
-      exitCode <- parseResult.fold(printHelp, identity)
+      exitCode <- parseResult.fold(printHelp[IO], identity)
     } yield exitCode
+  }
+
+}
+
+object CommandIOApp {
+
+  def run[F[_]](
+      name: String,
+      header: String,
+      helpFlag: Boolean = true,
+      version: Option[String] = None
+  )(opts: Opts[F[ExitCode]], args: List[String])(implicit F: Sync[F]): F[ExitCode] =
+    run(Command(name, header, helpFlag)(version.map(addVersionFlag(opts)).getOrElse(opts)), args)
+
+  def run[F[_]](command: Command[F[ExitCode]], args: List[String])(
+      implicit F: Sync[F]
+  ): F[ExitCode] =
+    for {
+      parseResult <- F.delay(command.parse(PlatformApp.ambientArgs getOrElse args, sys.env))
+      exitCode <- parseResult.fold(printHelp[F], identity)
+    } yield exitCode
+
+  private[CommandIOApp] def printHelp[F[_]: Sync](help: Help): F[ExitCode] =
+    Sync[F].delay(System.err.println(help)).as {
+      if (help.errors.nonEmpty) ExitCode.Error
+      else ExitCode.Success
+    }
+
+  private[CommandIOApp] def addVersionFlag[F[_]: Sync](
+      opts: Opts[F[ExitCode]]
+  )(version: String): Opts[F[ExitCode]] = {
+    val flag = Opts.flag(
+      long = "version",
+      short = "v",
+      help = "Print the version number and exit.",
+      visibility = Visibility.Partial
+    )
+
+    flag.as(Sync[F].delay(System.out.println(version)).as(ExitCode.Success)) orElse opts
   }
 
 }
--- a/effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
+++ b/effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
@@ -1,21 +1,28 @@
 package com.monovore.decline.effect
 
-import cats.effect.{IO, IOApp, ExitCode}
+import cats.ApplicativeError
+import cats.effect.implicits._
+import cats.effect.{Effect, ExitCode, IO, IOApp}
 import cats.implicits._
-
 import com.monovore.decline._
 
-abstract class CommandIOApp(
+import scala.language.higherKinds
+
+abstract class CommandIOApp[F[_]](
     name: String,
     header: String,
     helpFlag: Boolean = true,
     version: String = ""
-) extends IOApp {
+)(implicit ev1: Effect[F], ev2: ApplicativeError[F, Throwable])
+    extends IOApp {
 
-  def main: Opts[IO[ExitCode]]
+  def main: Opts[F[_]]
 
   private[this] def command: Command[IO[ExitCode]] = {
-    val showVersion = {
+    val mainCommand: Opts[IO[ExitCode]] =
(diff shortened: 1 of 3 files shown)


effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
```diff
@@ -1,21 +1,26 @@
...
+
+abstract class CommandIOApp[F[_]: Effect](
     name: String,
     header: String,
     helpFlag: Boolean = true,
     version: String = ""
 ) extends IOApp {
 
-  def main: Opts[IO[ExitCode]]
+  def main: Opts[F[_]]
```
johnynek: this destroys the ability of users to set exitcodes, no? This seems an orthogonal change.

andreamarcolin: Well, somehow I missed the public apply method for ExitCode and thought it was only possible to have ExitCode.Success and ExitCode.Error, which turns out to be wrong after a double check.
The original idea was not to force ZIO and Monix users to provide a cats-effect result (ExitCode), but the only other way I see to preserve the ability to explicitly set the exit code is via an Int. 
Using ExitCode is not that dramatic though, I think, so I am fine either rolling back to using it in the return type or using a plain Int, just let me know what you think about and I will push an update.

johnynek: if you are using this and ZIO, you are already depending on cats-effect (since this code does). I think using a type (ExitCode) is better than using Int. Not to mention, using `Int` would be a binary compatible (since this is erased) but could be a runtime error if users link against the wrong version.

I think using ExitCode here (not making a change) is far preferable.

andreamarcolin: Very good points! I pushed a fix

effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
```diff
@@ -1,21 +1,26 @@
...
+import cats.effect.implicits._
+import cats.effect.{Effect, ExitCode, IO, IOApp}
 import cats.implicits._
-
 import com.monovore.decline._
 
-abstract class CommandIOApp(
+import scala.language.higherKinds
+
+abstract class CommandIOApp[F[_]: Effect](
```
johnynek: actually, I guess mima must not be running, the `: Effect` will change the signature of this class to require the `Effect[F]` so I think this breaks binary compatibility.

I think it is better not to break binary compatibility personally. What about adding:
```scala
abstract class CommandEffectApp[F[_]: Effect]...
```
then make:
```scala
abstract class CommandIOApp(...) extends CommandEffectApp[IO](name, header,...)
```
That way, I don't think we break binary compatibility, but also ZIO users can use `CommandEffectApp[ZIO[X, ?]]` or whatever...

johnynek:
looks like #31 is still open...

effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
```diff
@@ -1,21 +1,26 @@
...
     version: String = ""
 ) extends IOApp {
 
-  def main: Opts[IO[ExitCode]]
+  def main: Opts[F[ExitCode]]
 
   private[this] def command: Command[IO[ExitCode]] = {
-    val showVersion = {
+    val mainCommand: Opts[IO[ExitCode]] =
+      main.map(_.toIO)
```
bkirwi: So FWIW this isn't exactly what I expected after our discussion on the issue. (Probably my fault!) I had understood it as wanting to use [zio.App](https://zio.dev/docs/overview/overview_running_effects) and its runtime system / etc. with `decline`. I'd been imagining some `def run[F[_] : Effect, A](command: Command[A], args: List[String]): F[A]` method that did the status-setting and error-printing, which could be used a la:

```scala
object MyApp extends zio.App {
  def command: ZIO[ZEnv, Nothing, Int] = ???
  def run(args: List[String]) = CommandIOApp.run(command, args)
}
```

I'm not hung up on that idea, though eg. it's not clear to me if there are downsides for ZIO people using IOApp (does the runtime stuff work out?)

Though in its current state this is mostly saving users a call to `.map(_.toIO)` themselves... which is not nothing but maybe not enough to justify the switch. Thoughts?

andreamarcolin: Well actually I didn't fully understood your idea until now, so definitely not your fault.
About using IOApp with ZIO runtime: when converting an Effect to IO, cats-effect runs the provided effect prematurely via unsafeRunSync, so breaking the idea of running effects only at "the end of the world". That is known to be non-ideal, but other (cats-effect based) major libraries do this at this  time (http4s) and it works to some extent.

About the proposed `def run[F[_]: Effect, A]`: could you please elaborate a bit more? In fact, I am still trying to wrap my head around your example. Are you suggesting to place it in the companion object? If yes, we would lose the showVersion part (due to already requiring Command and not Opts as a parameter) and need to replicate in a similar but polymorphic fashion the printHelp part. I can try to take at stab at it if this matches your idea, just wanted to be sure it is the way you want to go 😄

andreamarcolin:
> @johnynek - ah yeah, i should really get around to enabling that now that we're at `1.0`. 😅
> 
> @andreamarcolin - thanks for the followup on this! comments inline.

@bkirwi Somehow unrelated question: is it possible to mark 1.0.0 as 'latest release' in releases tab? I was fooled by the fact that 0.4.2 seems to be the latest release, so I didn't put the required attention to not break binary compatibility 😅 and obv merging this PR could be quite beneficial to catch problems like this. Ty!

effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
andreamarcolin: See my comment below, if you agree I would keep this as a backup solution in case we don't want to go with what @bkirwi proposed

andreamarcolin:
@bkirwi @johnynek I started from scratch ad tried to reimplement this keeping in mind your suggestions, approaching like @bkirwi suggested and avoiding breaking changes. Let me know what you think about this new version. Ty!

effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
```diff
@@ -14,34 +16,55 @@ abstract class CommandIOApp(
...
-    def printHelp(help: Help): IO[ExitCode] =
-      IO(System.err.println(help)).as {
-        if (help.errors.size > 0) ExitCode.Error
-        else ExitCode.Success
-      }
+}
+
+object CommandIOApp {
 
+  def run[F[_]](
```
johnynek: maybe expose `def command[F[_]]` and and just build the command up?

effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
```diff
@@ -14,34 +16,55 @@ abstract class CommandIOApp(
...
-          help = "Print the version number and exit.",
-          visibility = Visibility.Partial
-        )
-        flag.as(IO(System.out.println(version)).as(ExitCode.Success))
-      }
-    }
+  override final def run(args: List[String]): IO[ExitCode] = {
+    import CommandIOApp._
+
+    val opts = if (version.isEmpty) main else addVersionFlag(main)(version)
```
johnynek: why not call `CommandIOApp.run[IO](...)` here and remove the duplication?

andreamarcolin: The only little difference is the version parameter, being an Option in `CommandIOApp.run`, but it could be easily adjusted

effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
andreamarcolin: It would require two function calls in client code but I guess that's ok

johnynek: I think taking this suggestion is less important if you remove the duplication from the class and the companion object. I was looking for opportunities for reuse.

effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala
andreamarcolin: Pushed the removal of duplication 🙂

johnynek:
This looks good to me!

andreamarcolin:
@bkirwi gently pinging to know if everything is ok with this PR or if you want me to add/modify anything before merge 🙂

bkirwi:
out as 1.1.0 now if you want to give it a spin!

andreamarcolin:
Sorry to bother after that much time but, did it get released actually? I can't find 1.1.0 in the releases page, and also does not seem to show up on maven central. Thank you

bkirwi:
huh... true! looks like the release script failed partway... not sure what
happened there.

at this point it's probably smarter too go straight to 1.2.0. i'll try and
get that out tonight.

Le lun. 6 avr. 2020 à 07:07, Andrea Marcolin <notifications@github.com> a
écrit :

andreamarcolin:
Any news on this?

bkirwi:
Should be up now!

PR #126, opened 2019-12-30
Tweak error message

reibitto:
I know this is really minor so feel free to just close it if you don't like it, but I found the exclamation marks in the error messages kind of distracting, like:

```
Missing expected flag --rating!
```

It's jarring and it almost looks like it's part of the name.

I could just map the errors and strip them off, but I think the better default is to not include anything extra. Plus it's easier to append additional stuff than it is to strip stuff from strings (fewer corner cases to consider).

### DIFF
--- a/core/shared/src/main/scala/com/monovore/decline/Result.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/Result.scala
@@ -42,7 +42,7 @@ private[decline] object Result {
         if (envVars.isEmpty) None
         else Some(envVars.distinct.mkString("environment variable (", " or ", ")"))
 
-      s"Missing expected ${List(flagString, commandString, argString, envVarString).flatten.mkString(", or ")}!"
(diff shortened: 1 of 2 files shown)


bkirwi:
Hey, sorry, missed this one!

I guess the `!` is a little bit brazen, but I do value the punctuation. (A list of sentence fragments as errors can also be rather jarring.)

One way to do this would be to preserve the structure of the error in `Help`, so you could render it however you like. But that'd be a rather larger change...

reibitto:
Thanks. That kind of solution sounds good to me. I was also considering that because I was looking to add localization support and a "compact" mode (basically less newlines when rendering because the context I'm using this in is inside a window with limited space). If I have the data structure intact, I could add a custom render function to do either.

I'll close this one and create an issue for that. I'll eventually need it, so I wouldn't mind working on it. But I probably won't be able to get to it for a while. Maybe somebody else will get to it first.

PR #128, opened 2020-01-03
Add Show[Help] instance

chuwy:
IMO toString usage should be discouraged, but at the same time we would like to not break the compatibility.

If this is a right direction - I can change/add more tests.

### DIFF
--- a/core/shared/src/main/scala/com/monovore/decline/Help.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/Help.scala
@@ -1,5 +1,6 @@
 package com.monovore.decline
 
+import cats.Show
 import cats.data.NonEmptyList
 import cats.implicits._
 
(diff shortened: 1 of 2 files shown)


bkirwi:
hey, thanks for the pr!

i'm disinclined to deprecate the toString method, for a couple reasons:
- imo toString is fine in non-polymorphic contexts, and it’s more
accessible for folks who aren't super familiar with cats yet.
- there’s some demand for parametrized help rendering... if eg. you wanted
to set an indentation level for the rendered help, there’s nowhere obvious
to put that for a show instance. obviously not a concern yet! but we might
have to switch back to a method on help again soon anyways.

otoh, there’s no reason not to add the show instance, and it’s definitely
smart! i'd be quite happy to accept the additive parts of this.

chuwy:
Thanks @bkirwi! I still see a lot of value in getting rid of ubiqutous JVM methods, but I get that. I'll remove the deprecation and add some tests. Cheers!

chuwy:
Hey @bkirwi! I updated the PR.

bkirwi:
Great! Thanks in particular for the test.

I'll get this in as soon as I have some unrelated issues with Travis sorted out...

PR #134, opened 2020-01-23
Update discipline-scalatest

travisbrown:
There's a breaking change in the 1.0.0-RC4 release candidate, and it also transitively pulls in ScalaTest 3.1.0, which has some new deprecations.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -89,8 +89,8 @@ lazy val decline =
       description := "Composable command-line parsing for Scala",
       libraryDependencies ++= Seq(
         "org.typelevel"  %%% "cats-core"            % catsVersion,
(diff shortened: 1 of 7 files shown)


travisbrown:
The build failure is a 403 and looks like rate limiting or just general Maven Central weirdness, and should just need a restart from someone with permission.

bkirwi:
Hi Travis! Hope all is well.

Build's great after a restart. Useful if I get a release out with this soon?

travisbrown:
Thanks, Ben! I don't think we need a new release since this is a test dependency—I was just trying out the 1.0.0-RC4 upgrade on a handful of example projects that use discipline-scalatest before I published 1.0.0.

PR #145, opened 2020-02-12
List enumeratum choices in case of invalid argument

ivan-klass:


ivan-klass:
@bkirwi I believe there is something wrong with the code coverage project diff results, could you please take a look?
https://codecov.io/gh/bkirwi/decline/compare/273b855b40b4d0580db8a4ae13a84dc4acb70e12...f3fd57de0e320c8c747232a1216bb5a175b87f41/changes
May we proceed with this?
Appreciate your work. Thanks

enumeratum/src/main/scala/com/monovore/decline/enumeratum/package.scala
```diff
@@ -5,16 +5,23 @@ import cats.data.{Validated, ValidatedNel}
 import _root_.enumeratum._
 import _root_.enumeratum.values._
 
+import scala.collection.immutable.IndexedSeq
+
 package object enumeratum {
 
+  private[enumeratum] def invalidChoice(missing: String, choices: IndexedSeq[String]): String =
+    s"Invalid choice provided ($missing), choose one from <${choices.mkString(", ")}>"
```
bkirwi: This is stylistically a little bit out of line with the other error messages from this library; perhaps:

```suggestion
    choices match {
      case Seq() => s"Invalid value $missing"
      case Seq(value) => s"Invalid value $missing; expected $value"
      case many => s"Invalid value $missing; expected ${many.init.mkString(", ")} or ${many.last}"
    }      
```

enumeratum/src/main/scala/com/monovore/decline/enumeratum/ValueEnumArgument.scala
```diff
@@ -18,7 +18,7 @@ private[enumeratum] final class ValueEnumArgument[A, Entry <: ValueEnumEntry[A]]
       case Validated.Valid(value) =>
         enum.withValueOpt(value) match {
           case Some(r) => Validated.validNel(r)
-          case None => Validated.invalidNel(s"Invalid value: $string")
+          case None => Validated.invalidNel(invalidChoice(string, enum.values.map(_.value.toString)))
```
bkirwi: Had to check that `toString` is an inverse of parsing for all the basic value types, but that seems to be true!

ivan-klass: Thank you! I've also checked that. It maybe is better to require implicit Show here but I believe it would break binary compatibility

ivan-klass: Checked that it does not - ValueEnumArgument is private, and all the cases are listed in the package object.

bkirwi:
Merged -- thanks again!

PR #184, opened 2020-07-17
Add implicit Argument[Either[A, B]]

### LINKED ISSUE #182: combinators and typeclass instances on Argument
We could definitely have `Apply[Argument]` it seems (we don't have pure because can't supply a defaultMetavar) if we take the "left most" or "right most" defaultMetavar.

We can certainly have `Functor[Argument]` even without the above constraint.

But if we would allow the above (say take the left side). I think we can also have `Alternative[Argument]` which can be useful to compose different Arguments.

One thing I like is an
```scala
def either[A, B](a: Argument[A], b: Argument[B]): Argument[Either[A, B]] =
  Alternative[Argument].combineK(b.map(Right(_)), a.map(Left(_))) 
```

This allows a single flag like `--input` to parse multi values (say uris but also local file paths)

johnynek:
close #182

### DIFF
--- a/core/shared/src/main/scala/com/monovore/decline/Argument.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/Argument.scala
@@ -3,7 +3,7 @@ package com.monovore.decline
 import java.net.{URI, URISyntaxException}
 import java.util.UUID
 
-import cats.{Eval, Functor, SemigroupK}
+import cats.{Functor, SemigroupK}
 import cats.data.{Validated, ValidatedNel}
 
 import scala.annotation.implicitNotFound
@@ -16,7 +16,7 @@ import scala.annotation.implicitNotFound
 @implicitNotFound(
   "No Argument instance found for ${A}. For more info, see: http://monovore.com/decline/arguments.html#missing-instances"
 )
-trait Argument[+A] { self =>
+trait Argument[A] { self =>
 
   /**
    * Attempt to parse a single command-line argument: given an argument, this returns either
--- a/core/shared/src/main/scala/com/monovore/decline/Argument.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/Argument.scala
@@ -37,45 +37,33 @@ object Argument extends PlatformArguments {
 
   def apply[A](implicit argument: Argument[A]): Argument[A] = argument
 
+  /** convenience function to create Argument instances
+   */
+  def from[A](defMeta: String)(fn: String => ValidatedNel[String, A]): Argument[A] =
+    new Argument[A] {
+      override def read(string: String): ValidatedNel[String, A] = fn(string)
+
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/com/monovore/decline/Argument.scala
```diff
@@ -53,6 +77,41 @@ object Argument extends PlatformArguments {
...
+    new SemigroupK[Argument] {
+      override def combineK[A](x: Argument[A], y: Argument[A]): Argument[A] =
+        new Argument[A] {
+          override def read(string: String): ValidatedNel[String, A] = {
+            val ax = x.read(string)
+            if (ax.isValid) ax
+            else y.read(string)
+          }
+
+          override def defaultMetavar: String = s"${x.defaultMetavar}-or-${y.defaultMetavar}"
```
bkirwi, line 96: Thoughts on ` | ` instead of `-or-`? Usage will read like `--my-option <string | url>` vs. `--my-option <string-or-url>`.

Interested because pipe is already used in similar contexts in usage to express alternatives, it may read better if you end up with `a | b | c | d`, and users sometimes underscore their metavars which would make `-or-` a style mismatch.

core/shared/src/main/scala/com/monovore/decline/Argument.scala
```diff
@@ -36,6 +37,29 @@ object Argument extends PlatformArguments {
...
+      }
+
+    def mapValidated[B](function: A => ValidatedNel[String, B]): Argument[B] =
+      new Argument[B] {
+        def read(string: String): ValidatedNel[String, B] =
+          self
+            .read(string)
+            .andThen(function)
+
+        def defaultMetavar = self.defaultMetavar
```
bkirwi, line 59: One thing I worry about with these is preserving the metavar when it _should_ have changed. Let's say you do something like `Argument[String].mapValidated(parseInt)` or similar... you now have an `Argument` that only accepts ints but still has metavar `string`.

I'd be inclined to make this method take a new metavar alongside the function... WDYT? `map` seems safe because it can't reject anything, so from the CLI's perspective the range of valid inputs hasn't changed.

johnynek:
Note, typeclasses on typeclasses are quite common (they compose!), e.g. `InvariantMonoidal[Monoid]` `ContravariantMonoidal[Order]`...

core/shared/src/main/scala/com/monovore/decline/Argument.scala
johnynek: I agree that mapValidated keeping the defaultMetavar is a bit weird.

What I did here was add a `def from` method more compactly create Argument instances. I think that handles what I want from mapValidated anyway: I can just call read inside a function that I pass to `def from`.

johnynek:
Let me know what you think of these changes.

PR #185, opened 2020-07-17
Add Defer instance for Argument

johnynek:
Argument can have a defer instance, which allows use with some cats code that leverages Defer.


johnynek:
@bkirwi any concern with merging this?

johnynek:
I don't have an immediate use, I just like to provide instances for typeclasses in cats if we can so people can naturally work cats in expected ways. It is a speed bump when the instances which could exist are missing and then users have to write them in their own code and import them.

PR #192, opened 2020-08-28
Bump versions to support scalajs 1.x

johnynek:
There are a few problems:

1. we need cats 2.1.x to get scalajs 1.0
2. cats 2.1.x isn't published for 2.11 (which was released more than 6 years ago, BTW, and scala 3 is due out this year).

IMO we should merge this PR, but first do a final release that supports 2.11. I suspect very few users are still using 2.11, and if they are, they should be fine on old versions considering cats has dropped 2.11.


johnynek:
Copying from the cats build, I think this should work.

1. scoverage for scalajs 1.0 isn't supported (and it seems a bit redundant since the code is almost all the same).
2. the scalajs plugin only needs to be enabled when you are using scalajs namespace as far as I can tell. Very little of this code needs that.
3. like cats, we explicitly disable coverage for js so it won't fail to find the needed dependency.

johnynek:
What do you think of this one Ben?

Maybe release a 1.0.1 before this and then make this 1.1.0?

bkirwi:
Looks like the only merges since 1.2.0 are your recent PRs and some other version upgrades... I'll just go to 1.3.0 directly I think.

Thanks for your work on this! I'll ping here once the release is out.

bkirwi:
Seems like it just hit maven central at 1.3.0!

johnynek:
Awesome!

Thanks for the release!

PR #203, opened 2020-10-03
Allow help text rendering to be customized

felixbr:
Hi,

I've been using decline for my projects for a while now, thanks for the great library!

I've started using [the official Github cli](https://github.com/cli/cli) and it feels great to use. One of the things I like is the clear help text structure it has. Large cli applications like `gh` or `brew` structure their help text by topic rather than just list all options and flags and I think it's easier to find what you need (see screenshots below).

I tried to keep this change minimal. It just enables people like me to provide their own help text rendering function instead of the default one in decline. I know that you don't want to innovate on the bash side, so I hope this doesn't automatically disqualify.

Of course, I will change the PR however you like. For example I could call the function `Formatter` instead of `Printer`. I only used the latter because `circe` does so as well. 
Also if you feel this feature is too specific to be included in the arguments of `CommandApp` and `CommandIOApp`, I could also change it to a member of those classes, which can be overridden. That way it's less front and center.

Cheers
~ Felix

Screenshots of other CLIs that use a topic structure in their help text:

### DIFF
--- a/core/shared/src/main/scala/com/monovore/decline/CommandApp.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/CommandApp.scala
@@ -20,21 +20,17 @@ package com.monovore.decline
  * This should now behave like any other object with a main method -- for example, on the JVM, this
  * could be invoked as `java myapp.MyApp --fantastic`.
  */
-abstract class CommandApp(command: Command[Unit], helpPrinter: Help.Printer) {
-
-  def this(command: Command[Unit]) =
-    this(command, Help.defaultPrinter)
+abstract class CommandApp(command: Command[Unit]) {
 
   def this(
       name: String,
       header: String,
       main: Opts[Unit],
       helpFlag: Boolean = true,
-      version: String = "",
-      helpPrinter: Help.Printer = Help.defaultPrinter
+      version: String = ""
   ) {
 
-    this({
+    this {
       val showVersion =
         if (version.isEmpty) Opts.never
         else
@@ -43,8 +39,13 @@ abstract class CommandApp(command: Command[Unit], helpPrinter: Help.Printer) {
             .map(_ => System.err.println(version))
 
       Command(name, header, helpFlag)(showVersion orElse main)
-    }, helpPrinter)
+    }
   }
+
+  /**
+   * This method can be overridden to replace the default renderer with a custom one.
+   */
+  def renderHelp(help: Help): String = Help.defaultRenderer(help)
 
   @deprecated(
     """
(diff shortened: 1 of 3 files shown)


felixbr:
I'll address missing test coverage and documentation after initial feedback 🙂

bkirwi:
Hey - thanks for the thoughtful contribution, along with the links to prior art etc.! Just back from holiday so may not have a ton of time for this in the short term, but a few quick thoughts in case that's useful for now -

- You're right that we try to be pretty conservative in terms of the generated output etc. OTOH, I personally don't mind enabling folks to do less conventional things with it as long as it doesn't hurt the main use-case. (eg. https://github.com/tpolecat/tuco, which is not at all a conventional UNIX CLI but still gets some value from the library.)
- I also don't find this use case too farfetched, in terms of prior art; even Git does some grouping of commands. Confident I've seen tools that group options too, though none come immediately to mind...
- I don't think there's a need to name the function type here... any old function works fine!
- I'm a bit split on the change as implemented. The `*App` boilerplates are mostly there to help standardize... they're nothing you couldn't implement yourself, and the parameters take up as many lines as the job logic. In many cases, I'd rather folks copy-paste the few lines they need into their own projects instead of providing infinite configurability here. (Since that's at least a small burden on every user.) You're right that a member is less intrusive, though.
- Other libraries offer some more full-featured support here, including [one of the libraries that inspired this one](https://github.com/pcapriotti/optparse-applicative#command-groups). Significantly more complex change, but 
- In my use of the library I tend to use nested subcommands for a similar thing, grouping related commands together. Has some pros and cons, but worth considering even if you keep up with this approach.

I'll give this some more thought, but interested in your thoughts on any / all of the above.

felixbr:
Thanks for the feedback!

I'll make some changes to the current implementation to make it less intrusive. I was on the fence to use a class member anyway. Right now I have a slight preference for it, but we'll see how it turns out.

I've already written a wrapper around `CommandIOApp` to use `monix` with it instead but generally I don't like copy-pasting as a way to extend libraries. The setup around `cats-effect` is somewhat intricate regarding error-reporting, thread-pools and other configuration, so having a solid default setup is very valuable to many users imo.

The extensibility without copy-pasting also allows me and others to write a new renderer and publish it independently without having to copy a lot of your code (which then easily gets out-of-sync) or bothering you with PRs to include it directly.

I made the function very primitive (Help => String) because that way it doesn't prohibit any specific rendering choices. If we make the interface more restrictive (e.g. you can only customize the ordering and name of sections but not how they look like), it will prevent experimentation and we might have to make additional changes soon if someone wants to customize something we haven't anticipated. If I recall correctly, the helper functions the default rendering uses are public right now, so one can pick and choose what to re-use and what to write from scratch.

PR #206, opened 2020-10-18
Update cats-effect to 3.0.0-RC2

kubukoz:
Just checking if there will be any issues when CE3 actually comes out. Apparently not, since IOApp does most of the heavy lifting.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -69,7 +69,7 @@ lazy val noPublishSettings = Seq(
 )
 
 val catsVersion = "2.4.2"
-val catsEffectVersion = "2.3.1"
+val catsEffectVersion = "3.0.0-RC2"
 
 lazy val root =
   project.in(file("."))
--- a/effect/jvm/src/test/scala/com/monovore/decline/effect/CommandIOAppSpec.scala
+++ b/effect/jvm/src/test/scala/com/monovore/decline/effect/CommandIOAppSpec.scala
@@ -4,6 +4,7 @@ import cats.effect.ExitCode
 
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
+import cats.effect.unsafe.IORuntime
 
 class CommandIOAppSpec extends AnyFlatSpec with Matchers {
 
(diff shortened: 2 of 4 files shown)


kubukoz:
Apparently the build is a ScalaJS version issue.

bkirwi:
Thanks for checking!

And yep, not expecting any issues... our usage is very simple.

catostrophe:
Can we have this branch released as a milestone?

kubukoz:
cc @bkirwi @johnynek

johnynek:
Is there a way we can have the PR without the incidental reformatting? It makes it harder to understand the meat of the diff. I am not very opinionated about the style, but I'd really like to see the changes in two separate PRs.

kubukoz:
It's the style from .scalafmt.conf, I think the sbt files were just not formatted before... AFAICS the only relevant change in build.sbt is the version bump, so I can try to bring back just that one.

kubukoz:
That should do it.

kubukoz:
And this actually does.

bkirwi:
Looks reasonable! I should be able to cut a milestone in the next couple days.

I suppose this is a breaking change, so I guess that pushes us to 2.0?

kubukoz:
Yeah, I think so... also might want to add a version scheme: https://github.com/typelevel/cats/issues/3780

bkirwi:
Aight, this is out as `2.0.0-RC1`. Let me know if you run into any issues!

labra:
Thanks for the release. We tried it in the [shex-s project](https://github.com/weso/shex-s/) and it compiles and runs without arguments. However, when we try to run it with some arguments, it raises the exception: 

```
error] java.lang.NoClassDefFoundError: cats/effect/Sync
[error]         at es.weso.shexs.Main$.infoError(Main.scala:199)
[error]         at es.weso.shexs.Main$.$anonfun$main$3(Main.scala:195)
. . .
```

It seems there is some dependency on Cats-effec 2.0 but I don't know how to detect which one it is.

catostrophe:
@labra check `dependencyTree` in your project.

I have upgraded `decline` and can confirm that it doesn't pull CE2 dependency

labra:
Yes, indeed, I checked the dependencies again and the problem was that I had a dependency on `console4cats`. I removed it now and it works! Thanks for the update of decline to CE3!

kubukoz:
I do recommend upgrading to sbt 1.5.0-RC1 ASAP, if you have conflicting versions of cats-effect it'll break the build even before you compile! (thanks to the `versionScheme` setting)

PR #310, opened 2021-07-03
Add Duration plus FiniteDuration Argument

mdedetrich:
This PR adds `scala.concurrent.{Duration, FiniteDuration}` `Argument` readers for `Decline`. Note that the testing was a bit non trivial so I have comment the PR.

I have also verified that the tests pass locally.

### DIFF
--- a/core/shared/src/main/scala/com/monovore/decline/Argument.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/Argument.scala
@@ -7,6 +7,7 @@ import cats.data.{Validated, ValidatedNel}
 import cats.{Defer, Functor, SemigroupK}
 
 import scala.annotation.implicitNotFound
+import scala.concurrent.duration.{Duration, FiniteDuration}
 
 /**
  * This typeclass captures the information needed to use this type as an option argument.
@@ -179,6 +180,23 @@ object Argument extends PlatformArguments {
       try {
         Validated.valid(UUID.fromString(string))
       } catch { case _: IllegalArgumentException => Validated.invalidNel(s"Invalid UUID: $string") }
+    }
+
+  implicit val readDuration: Argument[Duration] =
+    from("duration") { string =>
+      try Validated.Valid(Duration(string))
+      catch { case _: NumberFormatException => Validated.invalidNel(s"Invalid Duration: $string") }
+    }
+
+  implicit val readFiniteDuration: Argument[FiniteDuration] =
+    new Argument[FiniteDuration] {
+      override def read(string: String): ValidatedNel[String, FiniteDuration] =
+        readDuration.read(string) andThen {
+          case fd: FiniteDuration => Validated.Valid(fd)
+          case _ => Validated.invalidNel(s"Invalid Duration: $string is not finite")
+        }
+
+      override def defaultMetavar: String = readDuration.defaultMetavar
     }
 
   /**
--- a/core/shared/src/test/scala/com/monovore/decline/ArgumentSpec.scala
+++ b/core/shared/src/test/scala/com/monovore/decline/ArgumentSpec.scala
@@ -1,9 +1,12 @@
 package com.monovore.decline
 
 import com.monovore.decline.discipline.ArgumentSuite
-import cats.{Defer, Eval, SemigroupK}
+import cats.{Defer, SemigroupK, Show}
 import cats.data.Validated
+import org.scalacheck.{Arbitrary, Gen}
+
 import java.util.UUID
+import scala.concurrent.duration.{Duration, FiniteDuration}
 
 class ArgumentSpec extends ArgumentSuite {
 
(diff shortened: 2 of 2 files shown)


core/shared/src/main/scala/com/monovore/decline/Argument.scala
```diff
@@ -181,6 +182,23 @@ object Argument extends PlatformArguments {
...
+  implicit val readDuration: Argument[Duration] =
+    from("duration") { string =>
+      try Validated.Valid(Duration(string))
+      catch { case _: NumberFormatException => Validated.invalidNel(s"Invalid Duration: $string") }
+    }
+
+  implicit val readFiniteDuration: Argument[FiniteDuration] =
+    new Argument[FiniteDuration] {
+      override def read(string: String): ValidatedNel[String, FiniteDuration] =
+        readDuration.read(string) andThen {
```
mdedetrich, line 194: Here we reuse the normal `Duration` reader however we make sure that its only finite using the `andThen`

core/shared/src/test/scala/com/monovore/decline/ArgumentSpec.scala
```diff
@@ -60,6 +63,53 @@ class ArgumentSpec extends ArgumentSuite {
   checkArgument[BigInt]("BigInt")
   checkArgument[UUID]("UUID")
 
+  val onlyValidArbitraryDuration = Arbitrary(
```
mdedetrich, line 66: The default `Arbitrary[Duration]` scalacheck property generator also generates `Duration.Undefined` which we don't care (there is no point in generating decline working with undefined `Duration`).

core/shared/src/test/scala/com/monovore/decline/ArgumentSpec.scala
```diff
@@ -60,6 +63,53 @@ class ArgumentSpec extends ArgumentSuite {
...
+    Gen.frequency(
+      1 -> Gen.const(Duration.Inf),
+      1 -> Gen.const(Duration.MinusInf),
+      1 -> Gen.const(Duration.Zero),
+      6 -> Gen.finiteDuration
+    )
+  )
+
+  val infinity1Show = Show.show[Duration] {
```
mdedetrich, line 75: If you check https://github.com/scala/scala/blob/8a2cf63ee5bad8c8c054f76464de0e10226516a0/src/library/scala/concurrent/duration/Duration.scala#L61-L63 you will see that there are multiple ways that the default `Duration.apply(s: String)` can parse negative/positive Infinity. To deal with these I have just created different variations of `cats.show.Show`  corresponding to the infinities.

Ideally I would have liked to have a single `cats.show.Show` that just uses the seed from Scalacheck and depending on that seed generates one of the various serializations of positive/negative Infinity however afaik its not possible to access the seed at this point (maybe with some refactoring of tests this is possible but if so that would be for another PR).

core/shared/src/main/scala/com/monovore/decline/Argument.scala
```diff
@@ -181,6 +182,23 @@ object Argument extends PlatformArguments {
...
+          case fd: FiniteDuration => Validated.Valid(fd)
+          case _ => Validated.invalidNel(s"Invalid Duration: $string is not finite")
```
mdedetrich, line 196: Is this error message appropriate?

core/shared/src/main/scala/com/monovore/decline/Argument.scala
```diff
@@ -181,6 +182,23 @@ object Argument extends PlatformArguments {
...
+        }
+
+      override def defaultMetavar: String = readDuration.defaultMetavar
```
mdedetrich, line 199: Not sure if you want to have `finite-duration` here instead of just the original `duration`?

mdedetrich:
@bkirwi Let me know if the PR needs any more additional work.

core/shared/src/test/scala/com/monovore/decline/ArgumentSpec.scala
```diff
@@ -1,9 +1,12 @@
 package com.monovore.decline
 
 import com.monovore.decline.discipline.ArgumentSuite
-import cats.{Defer, Eval, SemigroupK}
```
mdedetrich, line 4: `Eval` wasn't actually used so I just removed it from here.

core/shared/src/main/scala/com/monovore/decline/Argument.scala
bkirwi: Seems reasonable!

core/shared/src/main/scala/com/monovore/decline/Argument.scala
bkirwi: I think just `duration` is fine here as well.

(I try to think of these things from the perspective of the CLI user who has no idea how it's implemented internally, and it's not clear that they'd know or care about the distinction in this case... as a user I'd probably assume that all durations were finite unless it was called out explicitly in the docs.)

core/shared/src/test/scala/com/monovore/decline/ArgumentSpec.scala
bkirwi: Personally I'd also be comfortable with using the default `Show`/`toString` for all durations, and then adding a couple extra explicit test cases to cover the missing parts of the space. (As you can imagine this same issue crops up for floats as well... lots of ways to write down the same floating point number!)

Happy to keep this implementation too... but in that case, mind adding the context you shared here as a comment on the code? Might not be obvious to future readers why this is necessary.

bkirwi (commented):
This looks good... thanks for the contribution!

I checked the `Duration.apply` implementation, and it seems relatively friendly / intuitive even to folks who aren't familiar with Scala, so it does seem pretty reasonable to rely on it here.

One request on the tests but otherwise I think this is good to go.

core/shared/src/test/scala/com/monovore/decline/ArgumentSpec.scala
mdedetrich: So the issue is that the default cats `Show` for `Duration` does not work because the `String` values it outputs for the various infinite representations are different from what `Duration.apply(s: String)` accepts.

Are you fine with me going with the option of commenting code to provide context instead?

bkirwi: > So the issue is that the default cats Show for Duration does not work because the String values it outputs for the various infinite representations are different from what Duration.apply(s: String) accepts.

Right. In that case a `Show` that generates _any_ of the valid inputs for infinity would work here. (Note that we're not generating all of the possible valid inputs for finite durations either... `ms` vs. milliseconds etc.)

> Are you fine with me going with the option of commenting code to provide context instead?

Yep, also totally fine!

mdedetrich: Okay, I will create a single generic `Show` for `Duration` and then test the other infinite string representations explicitly, that should make it more clear

mdedetrich:
@bkirwi As discussed I have applied all of the changes.

core/shared/src/test/scala/com/monovore/decline/ArgumentSpec.scala
mdedetrich: Okay done, let me know if you are happy with it

bkirwi (approved):
Thanks for the followups!

PR #311, opened 2021-07-03
Move to Github Actions

bkirwi:
TravisCI is finally dead! Let's see how well Actions works.


bkirwi:
This appears to be a minimum viable replacement... will come back to it to add coverage etc. later.

PR #312, opened 2021-07-03
Update and apply .scalafmt.conf

mdedetrich:
This PR both updates the `.scalafmt.conf` so its compatible with the latest version of scalafmt and also applies it to the project (I noticed when working on my past PR that many files weren't being formatted likely due to forgetfullness and no check inplace to make sure files are formatted).

As is evident in the PR, scalafmt has changed a bit since 2.4.1 and hence a lot of the default formatting is different. @bkirwi let me know if you don't agree with some of the formatting and I can see if I can fix it via additional `.scalafmt.conf` options.

~~I would also recommend after merging this PR to add a github action to check if the PR is correctly formatted, this can be done using the sbt scalafmt plugin or by using scalafmt cli with the `--test` flag (which returns 1 if the code is not formatted).~~ This has already been done

Personally I find the `align.preset=most` most visually appealing however this is subject preference but let me know if you want updates to the formatting.


build.sbt
```diff
@@ -192,17 +199,17 @@ lazy val doc =
       micrositeDocumentationUrl := "usage.html",
       micrositeTheme := "pattern",
       micrositePalette := Map(
-        "brand-primary"     -> "#B58900",
```
mdedetrich, line 195: This is an example of formatting that doesn't look nice because the current formatting is `align.preset=none`.

mdedetrich:
I have just rebased this PR off master. Also I have added a github action that uses scalafmt to check that all code has been formatted when a PR is made (see `Scalafmt / Code is formatted`).

It uses scalafmt-native (so it runs really fast, usually under 10 second) and its also a separate pipeline which means it runs in parallel with the main build

build.sbt
bkirwi: Definitely less aesthetic, but it does have the advantage of shrinking diffs... personally fine with me.

bkirwi (approved):
Probably wise... thanks for tackling this!

PR #316, opened 2021-07-12
Add scala-native support

### LINKED ISSUE #42: Build for Scala Native?
What would be the complexity for providing decline for Scala Native?
I've seen that it depends on refined, which I do not understand clearly if would be available for Scala Native or not.
It also depends on ScalaTest, which in general can be easily replaced by uTest.
It also depends on ScalaCheck, which I simply would disable in the case of Scala Native.

Can you provide some guidance on how it could be ported?

Thanks

avdv:
* update Scala 2.12 to 2.12.13 since scala-native's nscplugin is available for
  2.12.13 and 2.12.14
* exclude java.nio.time support for scala-native, since scala-native does not
  implement this on its own and `org.ekrich %%% sjavatime` is not complete yet

Fixes #42

As already mentioned above, the scala-native module does not support the java.nio.time based Argument instances. I guess that's OK for a start, and could be improved later. Partial support would be possible, or one could separate the library part into a decline-time module which would not be published for scala-native at all.


build.sbt
```diff
@@ -115,6 +115,23 @@ lazy val decline =
       libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % "2.3.0",
       coverageEnabled := false
     )
+    .nativeSettings(
+      // Note: scala-native does not include a java.nio.time implementation and
+      //       the implementation from https://github.com/ekrich/sjavatime
+      //       is not complete
+      //       (missing 7 definitions while linking -- 16 without sjavatime)
+      libraryDependencies += "org.ekrich" %%% "sjavatime" % "1.1.5",
```
bkirwi, line 123: IIUC, this is both adding the dependency and excluding the code that will eventually use it... is that right?

(If so, maybe we should just add the new dep to the comment for now?)

bkirwi (commented):
Very cool! Kicking off the build.

Curious what you're using this for, if you don't mind sharing?

bkirwi:
Hmm, looks like CI will need an update... could you run `sbt githubWorkflowGenerate` and commit the result?

avdv:
> Curious what you're using this for, if you don't mind sharing?

Sure :smile, I have a little [side-project](https://github.com/avdv/scalals) where I am using scopt currently which comes with some idiosyncrasies and I'm thinking of replacing it with decline.

build.sbt
avdv: Yes, that's right. I'll comment this out for now.

avdv:
I've rebased onto master and pushed.

I would like to run the tests for Scala Native on CI too, but only for Scala 2.12 and 2.13. Do you know how to achieve that using sbt-github-actions?

bkirwi:
Hrm, I do not. The README makes it sound like it may be difficult: https://github.com/djspiewak/sbt-github-actions#generative-1

I suppose one might be able to reconfigure some of the tasks, so that eg. `githubWorkflowBuild` is a noop iff you're compiling on Scala 3, or look into some of the build matrix features? We'll also need to think about keeping `sbt release` working.

avdv:
> I suppose one might be able to reconfigure some of the tasks, so that eg. `githubWorkflowBuild` is a noop iff you're compiling on Scala 3, or look into some of the build matrix features?

I took a look at how they do it for cats. They have another matrix dimension:

```
ThisBuild / githubWorkflowBuildMatrixAdditions +=
  "platform" -> List("jvm", "js", "native")
```
and then exclude the Scala 3 / native platform job:
```
ThisBuild / githubWorkflowBuildMatrixExclusions +=
  MatrixExclude(Map("platform" -> "native", "scala" -> Scala3))
```
Also, they redefine the build steps and conditionally run with `if: matrix.platform == 'native'`.

I'll look into this a bit more tomorrow.

avdv:
I have again rebased onto master and added a job step which runs conditionally if the matrix.scala version is not Scala 3.

I'll look into getting `sbt release` working next.

bkirwi:
Kicked off the updated jobs - looks like everything's working smoothly.

avdv:
Sorry for responding so sluggishly... I have again rebased and pushed a new commit which also runs the publish artifacts step for the supported scala versions of the declineNative project.

Note, that the `runClean` and `runTest` release steps do not run the corresponding tasks for `declineNative`. It would be possible to integrate them in the same way as was done for the publish artifacts.

FTR, I tried aggregating the native project in the root project too, but that did not work so well since running any task with `+` always runs the tasks of the aggregated projects too, it just does not change the Scala version of the subproject. Which means that it would run `declineNative / test` with the Scala version set to the default one, even if it was switched to 3.0.1 globally. That would also lead to double publishes when running `release cross`.

bkirwi:
Cool looks good!

> Note, that the runClean and runTest release steps do not run the corresponding tasks for declineNative. It would be possible to integrate them in the same way as was done for the publish artifacts.

That sounds like a good idea, but not a blocker. (There's a risk that a compilation bug in Scala Native could fail the build fairly late... but, hopefully, not a large one.)

> Sorry for responding so sluggishly...

No need to apologize for volunteering your time slightly later than you expected to. :)

I'll go ahead and merge this, but will leave the corresponding issue open until we've got this released and confirmed working well.

avdv:
Thanks for merging! I will check this out for my project later on, but I am on vacation the next weeks... :smiley:

PR #321, opened 2021-07-15
Enable scoverage in github actions and add coveralls plus badge

mdedetrich:
This PR does the following things

1. Enables scoverage so it actually runs in CI builds
2. Adds coveralls (which works in conjunction with scoverage)
3. Adds a badge which shows the status of the coverage of the project. Note that it takes roughly a day for coveralls start displaying the coverage for a new project


build.sbt
```diff
@@ -12,8 +12,24 @@ lazy val Scala3 = "3.0.1"
...
+    name = Some("Upload coverage data to Coveralls"),
+    env = Map(
+      "COVERALLS_REPO_TOKEN" -> "${{ secrets.GITHUB_TOKEN }}",
+      "COVERALLS_FLAG_NAME" -> "Scala ${{ matrix.scala }}"
+    )
+  )
+)
 ThisBuild / githubWorkflowArtifactUpload := false
 ThisBuild / githubWorkflowPublishTargetBranches := Seq()
+// This is causing problems with env variables being passed in, see
```
mdedetrich, line 31: I am not sure why you specifically disable sbt thin client but in my experience I had to disable it because it doesn't pass ENV variables (which is needed for coveralls) so I just added a comment here explaining why

mdedetrich:
Hmm, so it seems that this is not working because scoverage doesn't support Scala3 yet. Theoritically its possible to put conditionals so that scoverage/scoveralls is only run for Scala3 but since this is non urgent I think just waiting for Scala3 support in scoverage is better.

mdedetrich:
Upstream issue at https://github.com/scoverage/scalac-scoverage-plugin/issues/299

PR #322, opened 2021-07-19
Also allow Duration argument to parse "infinity"

mdedetrich:
This PR allows the ability to parse `"infinity"` as a valid value for `Duration`. The primary reason for doing this change is that I noticed that akka libraries use "infinity" as a string value to designate `Duration.Inf` (i.e. https://github.com/akka/alpakka-kafka/blob/eb5246d3e03992ce60a1f33fefe75dadc58bc866/core/src/main/scala/akka/kafka/internal/ConfigSettings.scala#L44-L47)


bkirwi:
Shouldn't hurt!

PR #331, opened 2021-08-10
Add support for java.time.temporal.ChronoUnit

mdedetrich:
PR is quite self explanatory, adds support for `java.time.temporal.ChronoUnit` (only for the JVM since its not available on other platforms).

### DIFF
--- a/core/jvm/src/main/scala/com/monovore/decline/PlatformArguments.scala
+++ b/core/jvm/src/main/scala/com/monovore/decline/PlatformArguments.scala
@@ -1,8 +1,9 @@
 package com.monovore.decline
 
-import java.nio.file.{InvalidPathException, Path, Paths}
-
 import cats.data.{Validated, ValidatedNel}
+
+import java.nio.file.{InvalidPathException, Path, Paths}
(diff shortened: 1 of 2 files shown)


bkirwi:
Looks great on the technical side!

One principle I try to stick to for these is - would the output make sense to someone using the app who _isn't_ familiar with java / scala / the implementation / programming in general? (That why we eg. use `integer` instead of `long`.)
- Not sure about `chronounit` - what do you think about something like `time-units` instead?
- The values seem fine from this perspective... not sure about `half_days`, but that seems like a weird one to start with.

mdedetrich:
> Not sure about chronounit - what do you think about something like time-units instead?

Fixed

> The values seem fine from this perspective... not sure about half_days, but that seems like a weird one to start with.

Agreed, however at the same time I think design wise it makes sense to allow all `ChronoUnit`s and if someone wants a subset of values that they seem reasonable they can easily compose ontop of `readChronoUnit` to filter only a valid subset of values (in fact for my specific case I will be doing this as well but since this is meant to be a library I am trying to be general).

Issue #335, opened 2021-08-15
Should the values for ChronoUnit be kebab case?


mdedetrich:
After discussion at https://github.com/bkirwi/decline/pull/331 for adding `ChronoUnit` to pureconfig, a good point was made in that rather then the appropriate values for `ChronoUnit` being snake case they should instead be kebab case (or maybe something else more appropriate). The value that this effects is

```java
HALF_DAYS("HalfDays", Duration.ofSeconds(43200))
```

Currently with my PR at https://github.com/bkirwi/decline/pull/331 we accept either `HALF_DAYS` or `half_days` but maybe another format such as `half-days` is more appropriate.

There is also a discussion whether rather than accepting both `HALF_DAYS` and `half_days` that instead we should only accept a single standard format.

bkirwi:
I don't have strong feelings on this one... _leaning_ towards being as generous as possible in what we accept, since there's no clear winner, but open to the other approaches as well.

mdedetrich:
Will close this for now, if we decide to do something more fancy like whats done with pureconf a separate generic issue should be made about it

PR #353, opened 2021-09-11
Accept `half-days` for ChronoUnit Argument

mdedetrich:
This is a quick solution for https://github.com/bkirwi/decline/issues/335 which just replaces `_` with `-` so that `half-days` is an acceptable command line argument (as well as `half_days`)

There is an argument that we should do a nicer solution like what is done in pureconf, i.e. support for converting from one case to another (and vice versa) but this is a bigger task.

### DIFF
--- a/core/jvm/src/main/scala/com/monovore/decline/PlatformArguments.scala
+++ b/core/jvm/src/main/scala/com/monovore/decline/PlatformArguments.scala
@@ -24,7 +24,7 @@ private[decline] abstract class PlatformArguments {
 
     override def read(string: String): ValidatedNel[String, ChronoUnit] =
       try {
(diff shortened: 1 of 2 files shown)


bkirwi:
Thanks! Agree this is the pragmatic choice.

mdedetrich:
@bkirwi Would it be possible to make a release sometime soon?

bkirwi:
apologies, was away on holiday over the last couple weeks... i'll get one out in the next ~day.

bkirwi:
just released as 2.2.0, though it may still be percolating... let me know if you run into any issues!

mdedetrich:
Thank you very much!

PR #354, opened 2021-09-11
Update scalafmt

mdedetrich:
This PR updates scalafmt. Note that the new scalafmt also correctly formats docstrings hence the reason for the diff.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,5 +1,5 @@
-version=2.7.5
-docstrings=JavaDoc
+version=3.0.3
+docstrings.style=asterisk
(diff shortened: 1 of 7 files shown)


.scalafmt.conf
```diff
@@ -1,5 +1,5 @@
-version=2.7.5
-docstrings=JavaDoc
+version=3.0.3
+docstrings.style=asterisk
```
mdedetrich, line 2: Asterisk is the new name for the Javadoc style, see docstring in https://scalameta.org/scalafmt/docs/configuration.html

PR #370, opened 2021-11-17
Update/sbt scala native 0.4.1

avdv:
This supersedes #366 which only failed since the corresponding nscplugin for Scala 2.13.5 was not yet available on maven central.

Also, this unblocks #368 since scala-native 0.4.1 has been published for Scala 2.13.7.


bkirwi:
Thanks!

PR #380, opened 2021-12-14
Update docs for scala native & design advice

bkirwi:
A quick note on scala native support, which is currently undocumented, and a longer post with advice on structuring a large CLI.

The latter needs a bit more work; posting early in case anyone has thoughts on it!


doc/src/main/tut/structure.md
```diff
@@ -0,0 +1,222 @@
...
+Through every step of the refactoring above,
+the shape of our command-line parser closely matches the `Config` datastructure;
+we build parsers for case classes using `mapN` and sealed traits using `orElse`,
+working our way up from simpler types to the full `Config`.
+(And once that's done,
+we hand the whole thing off to some `run` function
+that does the actual work.)
+This "config" pattern is a very common way to structure an application using `decline`,
+and it's easy to test:
+you might write unit tests that pass different arguments to the parser
```
Daenyth, line 192: One testing pattern I have found value in is to just have a `List[String]` of all CLI invocations that are important to not regress, and just foreach assert that they don't result in a parse error. 

An example of this is if you use a command in cron, or something similar. This can help prevent it from being broken accidentally, and it's much lighter weight to just add the one invocation string to the "doesn't fail" list than a deep test. Often that's all you need

bkirwi: Yeah, good call -- I've seen that pattern work well too. Added a quick mention, but may expand more as I edit...

bkirwi:
Okay, I think this is in good enough shape to publish. I'll probably come back to it sooner or later, but if anyone else has thoughts feel free to open an issue or come chat in Discord.

Issue #404, opened 2022-03-31
CommandsApp exit code 0 when parsing fails


daddykotex:
Thanks for decline, it's truly amazing.

We use it as part of a few of our applications and I just realized that when using `CommandApp`, if the parsing fails, we print the help but still return 0 as the exit code. See https://github.com/bkirwi/decline/blob/master/core/shared/src/main/scala/com/monovore/decline/CommandApp.scala#L53

We had not realized before today because the `CommandIOApp` returns a different exit code when parsing fails: https://github.com/bkirwi/decline/blob/master/effect/shared/src/main/scala/com/monovore/decline/effect/CommandIOApp.scala#L42

Would you be open to a PR to do `sys.exit(1)` instead?

daddykotex:
see https://github.com/bkirwi/decline/pull/405

you merge and release when you see fit. if you can ping when this is actually out, great, otherwise I left a comment in our code that points here so we'll come around if you forget.

PR #405, opened 2022-04-01
CommandApp exit with non-zero code on parse failure #404

daddykotex:
JS implementation from:
https://github.com/alexarchambault/case-app/blob/d1da1ac55e97ce49748b84b2a99aecbb2d8452cd/core/js/src/main/scala/caseapp/core/app/PlatformUtil.scala


daddykotex:
I've tested on the JVM and it works fine (manually). I'm not sure how I'd go at it for unit test (this will probably murder the JVM running the tests)

bkirwi:
Thanks! I've tagged with `breaking`, so it'll be easier to find when we're looking at a breaking release.

PR #412, opened 2022-05-02
Capture environment variables in Node

plmuninn:
In scala-js `sys.env` can be empty. It's why I propose to allow explicitly setting it in run method - it would also allow testing application from top to bottom


bkirwi:
Hey, thanks for the contribution! Some quick notes:
- This is a breaking change. We could make it not-breaking by adding a forwarding method with the old signature... worth considering.
- In Scala.js, we deal with missing arguments by grabbing things from node's `process` in `PlatformApp`. It sounds like the same approach may work here? If so, we should strongly consider it. (This might avoid the need for a breaking change at all, actually, depending how we want to do it.)
- Scalafmt is unhappy; `sbt scalafmt` should fix.

bkirwi:
Hey, thanks for the contribution! Some quick notes:
- This is a breaking change. We could make it not-breaking by adding a forwarding method with the old signature... worth considering.
- In Scala.js, we deal with missing arguments by grabbing things from node's `process` in `PlatformApp`. It sounds like the same approach may work here? If so, we should strongly consider it. (This might avoid the need for a breaking change at all, actually, depending how we want to do it.)
- Scalafmt is unhappy; `sbt scalafmt` should fix.

bkirwi:
(And let me know if it's at all unclear how the `PlatformApp` approach works.)

plmuninn:
@bkirwi I will update on this tomorrow - just letting you know - somehow I missed `PlatformApp` from documentation. I will check it out and got back to you.

plmuninn:
@bkirwi I did changes to use `PlatformApp`. Let me know what do you think

bkirwi:
Looking good at first glance, but I won't have time for a proper review until perhaps next week.

bkirwi:
Great! This looks like it will work, and seems to be done in a way that avoids any bincompat issues. (Though the PR name is perhaps no longer accurate?)

Could you confirm the manual testing you did / that the latest version has the behaviour you'd expect when env vars are set?

plmuninn:
@bkirwi Yes I can confirm - we are using this code in our project inside company - I just copy pasted code from this PR and overwrite paths in class path. Everything works as expected.  

Name of branch is wrong - do you want me to close this PR and open new from different branch?

Issue #350, opened 2021-09-07
Support optional option arguments


avdv:
The GNU getopt function supports options that have optional arguments, which means they could be used as flags.

An example is the `--color` option from GNU ls:

* by default -- if the option is not given -- it's value is `auto`
* when using `--color` (as a flag) it has the value `always`
* otherwise, you can specify a value explicitly: `--color=never`

bkirwi:
yep - there's lots of precedent for this, so it would be nice to support. (though as a matter of personal style, i find a separate flag like `--default-color` or whatever is usually more clear.)

the main risk here is around parsing ambiguities... eg. around handling short options where a string `-ab` could be either an argument passed to option `-a` or two flags `-a` and `-b`. but assuming we can come up with something that's unambiguous and consistent with the usual standards ([POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/) etc.) this would be a good inclusion.

i'm unlikely to work on this anytime soon myself, but happy to discuss here if anyone's interested in tackling the implementation.

bkirwi:
Addressed in #415!

PR #415, opened 2022-05-04
Support for options with optional arguments 

nikololiahim:
An attempt to solve https://github.com/bkirwi/decline/issues/350. 

For now, it is unclear how an option with optional arguments should be parsed in presence of other options.
For long options, I suggest the following strategy:
* If the next arguments is special, e.g. an option or `--`, this argument should not be consumed and the option with optional argument should be considered empty. 
* Otherwise, the next agument should be consumed.

As for short options, the behavior should be the following (from what I understood of the [POSIX Utitility specification, section 2.b](https://pubs.opengroup.org/onlinepubs/9699919799/)):
* if the option with an optional option-argument is the last in the string of flags, it should be consumed as empty.
* Otherwise, this option should consume the entire remaining string of flags as its option argument.

@bkirwi what do you think?

### DIFF
--- a/core/shared/src/main/scala/com/monovore/decline/Parser.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/Parser.scala
@@ -131,12 +131,10 @@ private[decline] case class Parser[+A](command: Command[A])
 
 private[decline] object Parser {
 
-  type FlagOrArg = Option[String]
-
   sealed trait Match[+A]
   case class MatchFlag[A](next: A) extends Match[A]
   case class MatchOption[A](next: String => A) extends Match[A]
-  case class MatchOptArg[A](next: FlagOrArg => A) extends Match[A]
+  case class MatchOptArg[A](next: Option[String] => A) extends Match[A]
   case object MatchAmbiguous extends Match[Nothing]
 
   object Match {
@@ -317,10 +315,8 @@ private[decline] object Parser {
     ) extends Accumulator[NonEmptyList[Option[String]]] {
 
       override def parseOption(name: Opts.Name) =
-        if (names contains name) Some(MatchOptArg {
-          case Some(value) => copy(reversedValues = Some(value) :: reversedValues)
-          case None => copy(reversedValues = None :: reversedValues)
-        })
+        if (names contains name)
+          Some(MatchOptArg(arg => copy(reversedValues = arg :: reversedValues)))
         else None
 
       override def parseSub(command: String) = None
--- a/core/shared/src/main/scala/com/monovore/decline/Parser.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/Parser.scala
@@ -52,14 +52,7 @@ private[decline] case class Parser[+A](command: Command[A])
         accumulator.parseOption(Opts.LongName(option)) match {
           case Some(MatchFlag(next)) => consumeAll(rest, next)
           case Some(MatchAmbiguous) => failure(s"Ambiguous option/flag: --$option")
-          case Some(MatchOptArg(next)) =>
-            rest match {
-              case Nil => evalResult(next(None).result)
-              case value :: rest0 => 
-                // TODO: if "value" is a flag or an option - consume nothing, advance with (value :: rest0)
-                //       if not - consume "value", advance with "rest0"
-                ???
-            }
+          case Some(MatchOptArg(next)) => consumeAll(rest, next(None))
           case Some(MatchOption(next)) =>
             rest match {
               case Nil => failure(s"Missing value for option: --$option")
@@ -84,9 +77,12 @@ private[decline] case class Parser[+A](command: Command[A])
                 case NonEmptyString(nextFlag, nextTail) => consumeShort(nextFlag, nextTail, next)
               }
 
-            case Some(MatchOptArg(next)) => 
-              // TODO: implement consumption logic for short options
-              ???
+            case Some(MatchOptArg(next)) =>
+              tail match {
+                case "" => Right((rest, next(None)))
+                case value => Right((rest, next(Some(value))))
+              }
+
             case Some(MatchOption(next)) =>
               tail match {
                 case "" =>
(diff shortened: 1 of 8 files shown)


core/shared/src/main/scala/com/monovore/decline/opts.scala
```diff
@@ -133,6 +134,33 @@ object Opts {
...
+        case Some(value) => arg.read(value).map(Some(_))
+      }
+
+  def optionalOptArgs[A](
+      long: String,
+      help: String,
+      short: String = "",
+      metavar: String = "",
+      visibility: Visibility = Visibility.Normal
+  )(implicit arg: Argument[A]): Opts[(Int, List[A])] =
```
bkirwi, line 156: What do you think of `NonEmptyList[Option[A]]` instead? (Preserves more ordering information, and it's a "tighter" type in the sense that users don't have to stress about invalid values like `(0, Nil)`.)

core/shared/src/main/scala/com/monovore/decline/opts.scala
```diff
@@ -133,6 +134,33 @@ object Opts {
     Single(Opt.Regular(namesFor(long, short), metavarFor[A](metavar), help, visibility))
       .mapValidated(Argument[A].read)
 
+  def optionalOptArg[A](
```
bkirwi, line 137: Definitely a nitpick, but I'm not sure about the name. (Mix of full words and abbreviations, some of which aren't used elsewhere.) I'm not sure what else I'd suggest, though: maybe `optionalOption`, or `optionFlag` or `flagOption` to suggest you're getting both option-like and flag-like behaviours?

bkirwi (commented):
Hey, thank you for the contribution! Some comments inline.

> For long options, I suggest the following strategy:

From testing a few commands I have locally that use optional arguments, like `ls --color` or some git subcommands, it seems like they always expect the argument to be provided with the `=` syntax. (As in `ls --color=always`.) I think I prefer that approach; aside from my strong bias towards "doing the standard thing", it seems like it avoids some ambiguity, and is basically consistent with the short option behaviour.

> As for short options, the behavior should be the following (from what I understood of the [POSIX Utitility specification, section 2.b](https://pubs.opengroup.org/onlinepubs/9699919799/)):

This sounds exactly right. (And thanks for pulling up the spec for it!)

> @bkirwi what do you think?

Overall looks really good! A couple minor suggestions inline.

The last important thing we haven't discussed is updating the usage text. It seems like we'll probably just be able to format this as `--$name[=<$metavar>]` and otherwise treat it like any other option or flag, but I haven't gone through all the details... let me know if you run into any trouble there.

nikololiahim:
Thank you for the quick reply! I more or less understand what to do now.

nikololiahim:
@bkirwi I've implemented the things we discussed + added some tests to ParseSpec and HelpSpec. What do you think?

nikololiahim:
Nevermind, forgot to update the Usage part.

bkirwi:
Great!

I won't have time to take a proper look until next week, unfortunately, but at a first glance things are looking good. (Thanks in particular for updating the tests!)

nikololiahim:
That's great to hear! I'm pretty sure I have missed some important edge case, so don't hesitate to ping me.

core/shared/src/main/scala/com/monovore/decline/Parser.scala
```diff
@@ -95,7 +104,7 @@ private[decline] case class Parser[+A](command: Command[A])
         accumulator
           .parseSub(arg)
           .map { result =>
-            result(rest).swap.map { _.withPrefix(List(command.name)) }.swap.flatMap(evalResult)
+            result(rest).leftMap { _.withPrefix(List(command.name)) }.flatMap(evalResult)
```
bkirwi, line 107: Good call... I think this dates back to an older Scala version where this API was unavailable.

core/shared/src/test/scala/com/monovore/decline/HelpSpec.scala
```diff
@@ -39,6 +53,10 @@ class HelpSpec extends AnyWordSpec with Matchers {
           |        Second option.
           |    --third <integer>
           |        Third option.
+          |    --flagOpt[=<string>]
```
bkirwi, line 56: Let's give one of these two new options a short form, to see how that renders here.

(Looking at the code, I suspect we'll get `-F[=<metavar>]`, where the correct rendering would be `-F[<metavar>]`, in which case there's a minor bug to fix. But regardless the test will be useful to prevent any future trouble!)

bkirwi (commented):
Very cool! One possible bug, but otherwise this seems thorough and sound.

core/shared/src/test/scala/com/monovore/decline/ParseSpec.scala
```diff
@@ -86,6 +86,82 @@ class ParseSpec extends AnyWordSpec with Matchers with Checkers {
...
+        .orNone
+
+      val all =
+        Opts
+          .flag(long = "all", help = "...", short = "a")
+          .orFalse
+
+      val repeatedFlagOpt = Opts
+        .flagOptions[String](long = "flag", help = "...", short = "f")
+        .map(_.toList)
```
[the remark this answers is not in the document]
nikololiahim: True, will update.

core/shared/src/test/scala/com/monovore/decline/HelpSpec.scala
nikololiahim: Hmm, let me check...

nikololiahim: I think it would also make sense to extend the UsageSpec.

nikololiahim: Just checked, you were right. Updated the implementation and tests in https://github.com/bkirwi/decline/pull/415/commits/fd75004574383159275a8a2179a05e7564fd94a2.

core/shared/src/test/scala/com/monovore/decline/ParseSpec.scala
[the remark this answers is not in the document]
nikololiahim: Updated in https://github.com/bkirwi/decline/pull/415/commits/3131ab349c5425d9af8536b3b8e543ebb750495e.

bkirwi:
The updates look good!

The last thing I need to do is check this for backwards compat. I _think_ this is compatible, in which case I can cut a new minor release with the change. If it's not, it might take slightly longer to merge... but I'll try and get it out soon in any case.

PR #428, opened 2022-07-07
Enable Scala 3 cross version for native target

lavrov:
This became possible with discipline-scalatest v2.2.0 which is available for native 0.4.x


bkirwi:
Thank you for this! Glad to have the build a little more uniform.

Note that I'm going to be away from the desk for the next couple weeks, so I may not get a release out right away.

PR #446, opened 2022-10-07
Be lazy in defaults

seanmcl:
This is useful, for example, if you want to allow a temp file as an argument, but if none is provided calling Files.createTempFile.

### DIFF
--- a/core/shared/src/main/scala/com/monovore/decline/opts.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/opts.scala
@@ -57,7 +57,7 @@ sealed trait Opts[+A] {
 
   def orElse[A0 >: A](other: Opts[A0]): Opts[A0] = Opts.OrElse(this, other)
 
-  def withDefault[A0 >: A](default: A0): Opts[A0] =
+  def withDefault[A0 >: A](default: => A0): Opts[A0] =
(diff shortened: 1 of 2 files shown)


bkirwi:
FWIW, this is a binary-breaking change to a fairly core API and will probably be hard to merge!

I went back and forth on this point originally; lazy evaluation is undeniably useful, but also I appreciate the cats philosophy that having it everywhere can make APIs unnecessarily complex and unpredictable.

Normally I like to mention workarounds when a PR can't be merged right away, but in this case the method implementation is short enough that it's probably obvious to everyone involved. :) Another popular approach is to use an effect system like `cats-effect`, of course, which sidesteps the issue in a different way.

seanmcl:
Wow, I wouldn't have thought making an argument lazy changed the binary signature, though it makes perfect sense in retrospect. I understand that laziness can add complexity, though forcing an evaluation in this case seems not ideal. Even [cats uses laziness for `orElse` and friends](https://github.com/typelevel/cats/blob/708cbaba2e6ac07a5ac1819b6c18b0c83a2faaba/core/src/main/scala/cats/syntax/applicativeError.scala#L97v). I'm fine closing this even if I still think it's the right default! Not a big enough deal to make a major version update obviously :)

Issue #452, opened 2022-11-03
Requesting `--help` should not return same exit status as not requesting help


adamretter:
Hi there, I very much like decline, so thank you :-)
I do think there could be a small problem with an assumption around exit codes made by decline. So I thought I would ask here...

Given the following example from your docs:

```scala
def main(args: Array[String]) = tailCommand.parse(args, sys.env) match {
  case Left(help) =>
    System.err.println(help)
    sys.exit(1)
  case Right(parsedValue) =>
    // Your program goes here!
}
```

I cannot determine the difference between invoking `myapp --help` (which is a valid command) or just `myapp` (where some mandatory arguments are missing).

I have checked this against some linux utilities, and it seems that the convention I have seen is that:
* `myapp --help` is a valid code and should print to stdout, and then return exit code 0
* `myapp` where there are missing mandatory arguments, could print an error (and/or help message) to stderr, and then return exit code 1.

It seems to me that to follow this convention we would need the decline API to make the help in the Left of the disjunction a bit more informative as to whether the user specified the `--help` flag or not, e.g. something more like:

```scala
case class Help(requested: Boolean, text: String)

...

def main(args: Array[String]) = tailCommand.parse(args, sys.env) match {
  case Left(help: Help) if help.requested =>
    println(help.text)
    sys.exit(0)
  case Left(help: Help) =>
    System.err.println("Missing required mandatory arguments")
    System.err.println(help.text)
    sys.exit(1)
  case Right(parsedValue) =>
    // Your program goes here!
}
```

Any thoughts on this please?

bkirwi:
Happily this info is already available via the API, if slightly indirectly!
The `Help` case class includes a list of errors -- since it needs to be
able to print them -- so it would suffice to replace your `help.requested`
with `help.errors.isEmpty`. (In general it's safer to check whether there
were any errors than whether the help flag was specified, in case eg. they
specified other flags that cause a parse error.)

I'd be open to merging a helper method that gave users a more intuitive
shorthand for `errors.isEmpty`, and/or pointing out this option in the
docs. I think it's a bit late to change CommandApp and friends, though, at
least before the next major release.

(Not the point of your example, but printing "Missing required mandatory
arguments" there is not quite right... `decline` can produce a wide variety
of helpful errors that aren't just missing-required-options. Best to let
the errors speak for themselves.)

bkirwi:
[...]

PR #453, opened 2022-11-04
Differentiate between the request for help, and the desire to give help

adamretter:
Closes https://github.com/bkirwi/decline/issues/452

### DIFF
--- a/doc/src/main/tut/usage.md
+++ b/doc/src/main/tut/usage.md
@@ -205,9 +205,17 @@ A main method that uses `decline` for argument parsing would look something like
 
 ```scala mdoc:to-string
 def main(args: Array[String]) = tailCommand.parse(args, sys.env) match {
+
+  case Left(help) if help.errors.isEmpty =>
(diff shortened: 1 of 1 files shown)


doc/src/main/tut/usage.md
```diff
@@ -205,9 +205,17 @@ A main method that uses `decline` for argument parsing would look something like
 
 ```scala mdoc:to-string
 def main(args: Array[String]) = tailCommand.parse(args, sys.env) match {
-  case Left(help) =>
+
+  case Left(help) if (!help.errors.isEmpty)=>
```
bkirwi, line 209: Minor, but: let's drop the parens, and switch the order of these two cases so we don't need the `!` either?

bkirwi:
Looks good... thank you for the contribution! I'll push this up to the doc site soon.

adamretter:
@bkirwi Thanks so much for the software and the speedy response. Kind regards.

Issue #461, opened 2022-11-16
Simple `CommandIOApp` hangs


armanbilge:
And I have no idea why 🤔 h/t @s5bug for reporting.

```scala
//> using lib "org.typelevel::cats-effect::3.4.0"
//> using lib "com.monovore::decline-effect::2.3.1"

import cats.effect._
import com.monovore.decline._
import com.monovore.decline.effect._

object Main extends CommandIOApp("foo", "Foo program") {
  override def main: Opts[IO[ExitCode]] = {
    Opts(IO(ExitCode.Success))
  }
}
```

armanbilge:
This does not hang.
```scala
object Main extends IOApp {

  def main: Opts[IO[ExitCode]] = {
    Opts(IO(ExitCode.Success))
  }

  override final def run(args: List[String]): IO[ExitCode] =
    CommandIOApp
      .run[IO]("foo", "Foo program", true, Option("").filter(_.nonEmpty))(main, args)

}
```

PR #462, opened 2022-11-16
Version bumps

armanbilge:
Fixes https://github.com/bkirwi/decline/issues/461.

I'm sorry, I'm not really sure what we did in Cats Effect 3.4.0 that is not backwards compatible. But it seems that publishing new artifacts compiled against the latest CE fixes the deadlock.


bkirwi (commented):
Just got caught up on the bug in discord... pretty wild! Thanks for the PR.

Issue #475, opened 2022-12-10
Opts.env withDefault produces empty braces in usage


ivan-klass:
Adding  `withDefault` to `Opts.env` results in confusing empty braces being rendered in usage.

## Example:

```scala

import cats.syntax.all.given
import com.monovore.decline.{CommandApp, Opts}

object EmptyBraces
    extends CommandApp(
      "test",
      "test",
      (
        Opts.option[Int]("number", "Print N times", metavar = "N"),
        Opts.env[String]("GREETING", "A greeting").withDefault("Hi"), // FIRST []
        Opts.env[String]("BODY", "The content"), // NO DEFAULT - OK
        Opts.env[String]("FOOTER", "Ending").withDefault("Best regards"), // SECOND []
        Opts.option[String]("name", "Who to greet")
      ).mapN { case (n, greeting, body, footer, name) =>
        println(s"$greeting $name\n$body\n$footer" * n)
      }
    )

```
Usage output:
```
Missing expected flag --number!
Missing expected environment variable (BODY)!
Missing expected flag --name!

Usage: test --number <N> [] [] --name <string>

test

Options and flags:
    --help
        Display this help text.
    --number <N>
        Print N times
    --name <string>
        Who to greet

Environment Variables:
    GREETING=<string>
        A greeting
    BODY=<string>
        The content
    FOOTER=<string>
        Ending

```

Note ` [] [] ` in `Usage:` - according to my tests, It seems like every `Opts.env(...).withDefault` produces an empty braces pair

bkirwi:
Fascinating - thanks for the report! I'll take a look.

bkirwi:
`2.4.1` is out with the fix. Thanks again for the report!

PR #476, opened 2022-12-11
Handle the empty-optional case correctly

bkirwi:
The `Usage` rendering code tries to find groups of options that can be rendered using the `[--foo | --bar]` syntax -- effectively, a `Sum` that contains at least one empty `Prod`. However, if it contains _only_ empty `Prod`s, it will render as `[]`, which is nonstandard and not useful.

Previously, this would never arise in practise -- you'd need to write some weird opts instances like `Opts("foo").withDefault("bar")` to trigger it. However, as pointed out in #475, now that we have `Opts.env` this can arise with the much more reasonable `Opts.env(...).withDefault(...)`.

The proposed fix is to just filter out cases where the list of alternatives in the braces is empty. I'll want to do a little more testing before I'm confident in it, though.

### DIFF
--- a/core/shared/src/main/scala/com/monovore/decline/Usage.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/Usage.scala
@@ -16,26 +16,6 @@ private[decline] case class Usage(opts: Many[Options] = Prod(), args: Many[Args]
 
 private[decline] object Usage {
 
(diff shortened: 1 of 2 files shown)


ivan-klass (approved):

bkirwi:
I was right to be suspicious - my first take was filtering out too much. Happier with the code and tests now.

Fixes #457.

PR #539, opened 2024-05-11
WIP: render colours in help

keynmol:
Primary goals:

1. Colored output (exact design TBD)

CleanShot 2024-05-11 at 12 41 18

2. Maintain compatibility

### DIFF
--- a/core/shared/src/main/scala/com/monovore/decline/Help.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/Help.scala
@@ -3,6 +3,8 @@ package com.monovore.decline
 import cats.Show
 import cats.data.NonEmptyList
 import cats.syntax.all._
+import com.monovore.decline.HelpRenderer.Plain
+import com.monovore.decline.HelpRenderer.Colors
 
 case class Help(
     errors: List[String],
@@ -34,6 +36,13 @@ object Help {
     Show.fromToString[Help]
 
   def fromCommand(parser: Command[_]): Help = {
+    fromCommand(parser, HelpRenderer.Plain)
+
+  }
+
+  def fromCommand(parser: Command[_], renderer: HelpRenderer): Help = {
+
+    val theme = Theme.forRenderer(renderer)
 
     val commands = commandList(parser.options)
 
@@ -41,26 +50,28 @@ object Help {
       if (commands.isEmpty) Nil
       else {
         val texts = commands.flatMap { command =>
-          List(withIndent(4, command.name), withIndent(8, command.header))
+          List(withIndent(4, theme.subcommandName(command.name)), withIndent(8, command.header))
         }
-        List(("Subcommands:" :: texts).mkString("\n"))
+        List((theme.sectionHeading("Subcommands:") :: texts).mkString("\n"))
       }
 
     val optionsHelp = {
-      val optionsDetail = detail(parser.options)
+      val optionsDetail = detail(parser.options, theme)
       if (optionsDetail.isEmpty) Nil
-      else ("Options and flags:" :: optionsDetail).mkString("\n") :: Nil
+      else (theme.sectionHeading("Options and flags:") :: optionsDetail).mkString("\n") :: Nil
     }
 
     val envVarHelp = {
-      val envVarHelpLines = environmentVarHelpLines(parser.options).distinct
+      val envVarHelpLines = environmentVarHelpLines(parser.options, theme).distinct
       if (envVarHelpLines.isEmpty) Nil
-      else ("Environment Variables:" :: envVarHelpLines.map("    " ++ _)).mkString("\n") :: Nil
+      else
+        (theme.sectionHeading("Environment Variables:") :: envVarHelpLines.map("    " ++ _))
+          .mkString("\n") :: Nil
     }
 
     Help(
       errors = Nil,
-      prefix = NonEmptyList(parser.name, Nil),
+      prefix = NonEmptyList.of(parser.name),
       usage = Usage.fromOpts(parser.options).flatMap { _.show },
       body = parser.header :: (optionsHelp ::: envVarHelp ::: commandHelp)
     )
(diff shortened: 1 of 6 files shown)


core/shared/src/main/scala/com/monovore/decline/HelpFormat.scala
```diff
@@ -0,0 +1,27 @@
...
+  def colorsEnabled: Boolean
+}
+object HelpFormat {
+  case object Plain extends HelpFormat {
+    override def colorsEnabled: Boolean = false
+  }
+  case object Colors extends HelpFormat {
+    override def colorsEnabled: Boolean = true
+  }
+  case class AutoColors(env: Map[String, String] = sys.env) extends HelpFormat {
```
bkirwi, line 13: Neat - was not aware of this!

This could be a method returning a format instead of a format of its own... and I'd prefer to not include `sys.env` as a default argument, since it's arguably not pure.

core/shared/src/main/scala/com/monovore/decline/HelpFormat.scala
```diff
@@ -0,0 +1,27 @@
+package com.monovore.decline
+
+sealed abstract class HelpFormat extends Product with Serializable {
```
bkirwi, line 3: I'd be tempted to define this on the `Help` companion so folks can refer to `Help.Format` and `Help.Plain` etc., but I don't feel so strongly.

core/shared/src/test/scala/com/monovore/decline/HelpSpec.scala
```diff
@@ -42,20 +42,28 @@ class HelpSpec extends AnyWordSpec with Matchers {
...
+        """
+          |Usage:
+          |    program [--first] [--second <integer>] [--third <integer>] [--flagOpt[=<string>]] --flag[=<string>] [--flag[=<string>]]... run
           |
           |A header.
           |
           |Options and flags:
           |    --first, -F
           |        First option.
+          |
```
bkirwi, line 57: Can we do this without updating the existing spacing?

core/shared/src/main/scala/com/monovore/decline/Theme.scala
```diff
@@ -0,0 +1,50 @@
+package com.monovore.decline
+
+import cats.Show
+import cats.data.NonEmptyList
+import cats.syntax.all._
+import com.monovore.decline.HelpFormat.Plain
+import com.monovore.decline.HelpFormat.Colors
+
+private[decline] trait Theme {
```
bkirwi, line 9: I don't follow why we need both `Theme` and `HelpFormat` - seems like users could pass in a `Theme` directly. (And if we can get away with fewer concepts, strong preference for that!)

core/shared/src/main/scala/com/monovore/decline/Help.scala
```diff
@@ -3,70 +3,130 @@ package com.monovore.decline
...
+        envHelp = collectEnvOptions(parser.options).distinct,
+        usages = Usage.fromOpts(parser.options),
+        description = parser.header
+      )
     )
+
   }
 
-  def optionList(opts: Opts[_]): Option[List[(Opt[_], Boolean)]] = opts match {
+  private[decline] case class HelpArgs(
```
bkirwi, line 120: Neat! This seems like something with this shape could eventually be cleaned up and made public for some of the fancier help-rendering ideas folks have had (HTML generation, etc.) but leaving it private for now seems like the right call.

bkirwi (commented):
Left a few comments, but overall on board!

Will still need to take some care with compatibility breaking, but I don't see any reason why we shouldn't be able to ship this in a compatible way...

core/shared/src/main/scala/com/monovore/decline/Theme.scala
keynmol: > seems like users could pass in a Theme directly

I treat `Theme` interface as almost an implementation details - the methods in it are very tightly coupled to the help rendering logic (they're by no means the minimal set of methods required to render something), and I don't think it's beneficial (at **this point**) to expose this interface to users at all - because it'd increase compatibility surface.

Alternatively I could make it all 1 concept and make every method in Theme private to address the concerns from previous paragraph - but I felt like keeping Theme separate and private is the easiest w.r.t. API surface and flexibility.

core/shared/src/test/scala/com/monovore/decline/HelpSpec.scala
keynmol: Yep, I just always felt the default is a bit too crowded so thought I'd sneak it in :P Will revert.

bkirwi: Yeah... I don't know if I disagree on an aesthetic level, but there are a lot of benefits to compactness in this context. (This change on its own makes this section ~50% larger, and now options get more spacing than section headers, which would probably want even more breathing room proportionately...)

core/shared/src/main/scala/com/monovore/decline/Theme.scala
bkirwi: Yeah, it's true that the fact that it's not user-facing does mitigate it to a large extent. Willing to keep it as is for now...

keynmol:
While the changes are binary compatible, they're not source compatible (which I believe is allowed for a minor release in our semver conventions) – I was in fact hoisted by my own petard in my own project, where `.copy(body = Nil)` is no longer possible: https://github.com/indoorvivants/sn-bindgen/blob/main/modules/bindgen/src/main/scala/main.scala#L17

core/shared/src/test/scala/com/monovore/decline/HelpSpec.scala
```diff
@@ -43,7 +43,9 @@ class HelpSpec extends AnyWordSpec with Matchers {
       }
 
       Help.fromCommand(parser).toString should equal(
-        """Usage: program [--first] [--second <integer>] [--third <integer>] [--flagOpt[=<string>]] --flag[=<string>] [--flag[=<string>]]... run
+        """
+          |Usage:
+          |    program [--first] [--second <integer>] [--third <integer>] [--flagOpt[=<string>]] --flag[=<string>] [--flag[=<string>]]... run
```
bkirwi: Got to change this back! (Among other things, the single-line usage statement is extremely standard in the wild.)

bkirwi (commented):
Yeah, wow, that does look like a hassle! Looks like it worked out, though - Mima does seem happy on my end also.

I am, nevertheless, going to request some minor changes. Sorry! Hopefully a fairly light lift, but let me know if you feel otherwise.

core/shared/src/test/scala/com/monovore/decline/HelpSpec.scala
keynmol: Reverted begrudgingly to the original format :) https://github.com/bkirwi/decline/pull/539/commits/d8658fbd2c62907023957785f714567ec6ec6eef

bkirwi (approved):

keynmol:
> I think going in this direction doesn't box us out of doing that in the future. (This API would just build on some more general API.)

Yeah, this would require de-casing the `HelpArgs` case class and all of the case classes it reference, with a re-work of their public API to make it usable.

This is something I'm also interested in (as I'd like to make the rendering of my [CLI options](https://sn-bindgen.indoorvivants.com/configuration/index.html#full-help) prettier for the website), but I'm unlikely to tackle that at the moment either.

PR #542, opened 2024-05-29
Expose `val command: Command[IO[ExitCode]]` in `CommandIOApp`

lolgab:
If you want to show the help when handling errors in the `IO` part of the app, you need to call `.showHelp` on the command instance. This is not possible with `CommandIOApp` since it doesn't expose the app but creates and uses it on the fly.
This PR adds a `val command: Command[IO[ExitCode]]` to `CommandIOApp` so you can have similar errors to the errors printed by `validate` but when handling IO errors in the app with `.handleErrorWith` in the application logic

### DIFF
--- a/effect/jvm/src/test/scala/com/monovore/decline/effect/CommandIOAppSpec.scala
+++ b/effect/jvm/src/test/scala/com/monovore/decline/effect/CommandIOAppSpec.scala
@@ -24,6 +24,18 @@ class CommandIOAppSpec extends AnyFlatSpec with Matchers {
     runApp() shouldBe ExitCode.Error
   }
 
+  it should "expose command so showHelp can be called" in {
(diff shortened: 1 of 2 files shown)


bkirwi:
Thanks for the contribution!

I suspect this is a breaking change, but I'll need to confirm. (Offhand this sort of thing is usually not enough to prompt a breaking release, though I do hang on to minorly-breaking PRs for whenever that time does come.) 

Generally the thinking has been that IO errors are not validation errors, and thus not the sort of thing you print help output for, so it was fine that it was annoying to overlap the layers in that way. If you've got a specific example where that's not the case I'd be interested to know about it though!

bkirwi:
Another way to solve this problem would be to mimic how `CommandApp` works, and add a second constructor which takes a command directly. Then you'd be able to construct it separately and thread it through wherever it's needed.

PR #543, opened 2024-06-18
WIP add configurable composable paring API

SpaceCowboyMax:
I've added some sort of public API for customize a Parser args consumer, while maintaining backward compatibility and preserving the default Parser logic. However, I still need to refine the handling of default implicits, naming things, and the final API design.

Proposed API - still WIP

```scala
implicit val customLongOptParser: ParserConfiguration[(String, String)] =
  ParserConfiguration(
    prepend = false,
    parserConfig = LongOptWithEqualsParser[(String, String)](pattern = "\\+(.+?)\\+(.+)".r) ::
      LongOptParser[(String, String)](skipUnknown = true) :: Nil
  )

val opts = (whatever, ghost).tupled
val cmd = Command("", "")(opts)

val result = cmd.parse(List("--whatever", "man", "+ghost+dad", "--unknownLongOpt"))
result should equal(Right("man", "dad"))
```

Feedback needed on topic: "Is it worth trying?"

I know the main idea of `decline`, but I hope this approach will provide some wiggle room.

### DIFF
--- a/core/jvm/src/test/scala/com/monovore/example/decline/Git.scala
+++ b/core/jvm/src/test/scala/com/monovore/example/decline/Git.scala
@@ -2,6 +2,7 @@ package com.monovore.example.decline
 
 import cats.syntax.all._
 import com.monovore.decline._
+import Command.defaultParserConfig
 
 object Git
     extends CommandApp(
--- a/core/jvm/src/test/scala/com/monovore/example/decline/ScoptExample.scala
+++ b/core/jvm/src/test/scala/com/monovore/example/decline/ScoptExample.scala
@@ -4,6 +4,7 @@ import java.nio.file.Path
 
 import cats.syntax.all._
 import com.monovore.decline._
+import com.monovore.decline.Command.defaultParserConfig
 
 object ScoptExample
     extends CommandApp(
--- a/core/shared/src/main/scala/com/monovore/decline/CommandApp.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/CommandApp.scala
@@ -37,7 +37,7 @@ abstract class CommandApp(command: Command[Unit]) {
           Opts
             .flag("version", "Print the version number and exit.", visibility = Visibility.Partial)
             .map(_ => System.err.println(version))
-
+      import Command.defaultParserConfig
       Command(name, header, helpFlag)(showVersion orElse main)
     }
   }
--- a/core/shared/src/main/scala/com/monovore/decline/Parser.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/Parser.scala
@@ -7,129 +7,26 @@ import cats.syntax.all._
 import com.monovore.decline.Opts.Name
 import com.monovore.decline.Parser.Accumulator.OrElse
 
-import scala.annotation.tailrec
 import scala.util.{Left, Right}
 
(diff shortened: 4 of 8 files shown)


bkirwi:
Thanks for the contribution!

> Feedback needed on topic: "Is it worth trying?"

Honestly, I'm pretty hesitant! It seems to fall outside the original scope of `decline`, expand the API in a pretty substantial way, and make error reporting etc. more difficult... and it's not totally clear to me what it's _for_.

Can you say a little more about the use-case that motivated this kind of change?

SpaceCowboyMax:
In my case, I would like to use decline for parsing a lot of non-deterministic lists of different options and arguments from files without any particular ordering and with gnu long_opt_only format via `--foo` which `-foo`. At first glance, it could be a good solution to write a custom parser, but with arg parsing, things could become very complicated and ugly. The same for input preprocessing. 

With such kind of API gnu getopt_long_only could be add and many other small things based on already pre-existed parsers
```scala
def gnuGetoptLongOnly[T](nonStrict: Boolean): List[ParserConfig[T]] =
  List(
    LongOptWithEqualsParser[T](skipUnknown = nonStrict, pattern = "-(.+?)=(.+)".r),
    LongOptParser[T](skipUnknown = nonStrict, pattern = "-(.+)".r)
  )

def skipCommentsParser[T]: ParserConfig[T] = new ParserConfig[T] {
  override val consumer: Consumer[T] = {
    case Consumable(arg :: rest, accumulator) if arg.startsWith("//") =>
      continue(Consumable(rest, accumulator))
  }
}

def parserConfig[T](nonStrict: Boolean): ParserConfiguration[T] =
  ParserConfiguration(
    parserConfig = skipCommentsParser[T] +: gnuGetoptLongOnly[T](nonStrict)
  )

val defaultCommand = command("default", opts)(Command.defaultParserConfig)
val strictCommand = command("strict", opts)(parserConfig(false))
val nonStrictCommand = command("non-strict", opts)(parserConfig(true))

defaultCommand.parse(
  List("--foo", "foo", "--bar", "bar", "buzz")
) shouldEqual Right("foo", "bar", "buzz")

strictCommand.parse(
  List("-foo", "foo", "//comment-this-out", "-bar", "bar", "buzz")
) shouldEqual Right("foo", "bar", "buzz")

nonStrictCommand.parse(
  List("-foo", "foo", "//comment-this-out", "-bar", "bar", "buzz", "-unkonw")
) shouldEqual Right("foo", "bar", "buzz")
```

bkirwi:
Got it, thanks!

It's a neat idea, and I can see where you're coming from, but I don't think it's a good fit for `decline`.

Thinking, in case it's interesting to you or some future reader:
- `decline` has historically limited itself to the minimum needed to implement a "normal" / "standard" UNIXy CLI. This is obviously annoying if you want to parse some other thing, but I like that it guides authors to write CLIs with very standard behaviour. (It's also a feature for me, the maintainer, since it keeps the scope of the library nice and small.)
    - `getopt_long_only` is interesting prior art -- thank you for mentioning it -- but it seems pretty rare in the wild.
- I believe adding flexibility here makes generating usage statements and error messages more difficult to do helpfully and accurately, or require more effort from the user to get good output, and high-quality usage and error messages are a core priority for this library.

I'm sorry - I know this isn't the answer you were hoping for! Best of luck with your project -

PR #544, opened 2024-06-20
Use cats.syntax.all instead of cats.implicits

lenguyenthanh:
follow this pr: https://github.com/typelevel/cats/pull/4394


bkirwi:
Thanks for the cleanup!

PR #558, opened 2024-08-28
Accumulate errors when reading Eithers

hugo-vrijswijk:
This is useful feedback when both sides of the Either are invalid. You want to know why both sides failed, not just the Right


hugo-vrijswijk:
@bkirwi could you take a look at this? 🙏

bkirwi:
I was a bit divided on this originally, since it breaks with `decline`'s usual error handling style... normally if we expect "A or B" we report that as a single error/line, not multiple lines. This is supposed to distinguish between having two different errors, or a single error that might be resolved two ways.

But I think following that principle here would involve reworking some interfaces here, and your PR solves a practical problem in a reasonable way! So I'll merge this in, and leave a note to revisit if we ever do a breaking release. Thanks for the PR, and the ping!

PR #577, opened 2024-12-14
Scala native 0.5, Scala 3 LTS and disable Cats Effect Native integration

keynmol:
Otherwise, only version bumps.


bkirwi:
Thanks for the contribution!

We're probably overdue for a release... I'll see if I can get one out in the next few days.

PR #589, opened 2025-02-14
Adds latest version and typelevel badges.

everson:
title says it all ;)

You can see it in https://github.com/everson/decline/tree/latest_version_badge?tab=readme-ov-file#decline


bkirwi:
Thank you! This will be very helpful for folks I think.

PR #618, opened 2025-09-10
Publish decline-effect to scala-native

hugo-vrijswijk:


bkirwi:
Exciting! I should be able to get this released in the next few days.

PR #645, opened 2026-02-09
Fix: autoColors works in the opposite way

keynmol:
Whoops.

I reworked this condition because `exists` and `nonEmpty` made my head hurt a little bit.

### DIFF
--- a/core/shared/src/main/scala/com/monovore/decline/HelpFormat.scala
+++ b/core/shared/src/main/scala/com/monovore/decline/HelpFormat.scala
@@ -133,7 +133,7 @@ object HelpFormat {
    * @return
    */
   def autoColors(env: Map[String, String]) =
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/com/monovore/decline/HelpFormat.scala
```diff
@@ -133,7 +133,7 @@ object HelpFormat {
    * @return
    */
   def autoColors(env: Map[String, String]) =
-    Plain.withColors(colors = env.get("NO_COLOR").exists(_.nonEmpty))
+    Plain.withColors(colors = env.get("NO_COLOR").map(_ => false).getOrElse(true))
```
bkirwi: It looks like the original intent was to have `NO_COLOR=''` have colors enabled... is that still desirable?

bkirwi (approved):
Thanks for the followup!

